PR #10, opened 2017-03-21
Check the version string of installed binary, ref #8

### LINKED ISSUE #8: One-line installer doesn't detect package installation failure
http://pastebin.com/MmmWeyRs

Says `Parity has been installed`

5chdn:
Should fix #8 

Needs testing on Ubuntu and MacOS.


5chdn:
Tested and confirmed working on Ubuntu 14.4 and 16.4. Installs correctly, or notices if installed version does not match, and aborts.

No mac available for testing:

    bash <(curl https://raw.githubusercontent.com/5chdn/scripts/master/install-parity.sh -Lk)

PR #16, opened 2017-07-29
Silence Curl

MysticRyuujin:
Script has become noisy. I propose adding -sS to the curl commands

-S, --show-error    Show error. With -s, make curl show errors when they occur
-s, --silent        Silent mode (don't output anything)


MysticRyuujin:
curl_noise

PR #18, opened 2017-10-13
Default to Stable

MysticRyuujin:
1.7.5 is the current recommended version and it is "stable"

Also, I'm not sure why it defaults to "beta" in the first place. Normal users should always be on a stable branch...


5chdn:
should default to beta for end-users. 1.8.0 beta will be available asap.

MysticRyuujin:
Our offline discussion suggests beta = stable haha I concede this pull request

PR #22, opened 2018-01-26
Add an image of the JS libp2p echo listener

tomaka:
Desirable for adding a test that checks whether our Rust port of libp2p can successfully connect to it.

The working directory handling is a bit of a clusterfuck. Let me know if I need to fix it.

cc https://github.com/libp2p/rust-libp2p/pull/112


General-Beck:
I will add then automatic assembly to docker hub

PR #23, opened 2018-01-26
Various updates for linux installer.

ddorgan:
1) Use debian package on debian derived systems instead of ubuntu package
2) No longer depend on the snap package being present from the vanity service
3) Attempt an rpm install if rpm is found (currently only a centos package is built).

### DIFF
--- a/get-parity.sh
+++ b/get-parity.sh
@@ -1,6 +1,5 @@
 #!/bin/bash
 # Copyright 2017 Parity Technologies (UK) Ltd.
-set -x
 RELEASE="beta"
 ARCH=$(uname -m)
 VANITY_SERVICE_URL="https://vanity-service.parity.io/parity-binaries?architecture=$ARCH&format=markdown"
@@ -48,7 +47,7 @@ check_os() {
 
 		if [ "$DISTRO" = "jessie" ] || [ "$DISTRO" = "wheezy" ] || [ "$DISTRO" = "stretch" ] ; then
 			PKG="debian"
-		elif [ "$DISTRO" = "xenial" ] || [ "$DISTRO" = "trusty" ]  || [ "$DISTRO" = "precise" ] ; then
+		elif [ "$DISTRO" = "xenial" ] || [ "$DISTRO" = "trusty" ] || [ "$DISTRO" = "precise" ] || [ "$DISTRO" = "artful" ] || [ "$DISTRO" = "bionic" ] ; then
 			PKG="linux"
 		fi
 	      fi
--- a/get-parity.sh
+++ b/get-parity.sh
@@ -1,19 +1,61 @@
 #!/bin/bash
 # Copyright 2017 Parity Technologies (UK) Ltd.
+set -x
 RELEASE="beta"
 ARCH=$(uname -m)
 VANITY_SERVICE_URL="https://vanity-service.parity.io/parity-binaries?architecture=$ARCH&format=markdown"
+DISTRO=$(lsb_release -c -s)
+
+check_alt() {
+    if [ "X${DISTRO}" == "X${2}" ]; then
+        echo
+        echo "## You seem to be using ${1} version ${DISTRO}."
+        echo "## This maps to ${3} \"${4}\"... Adjusting for you..."
+        DISTRO="${4}"
+    fi
+}
 
 check_os() {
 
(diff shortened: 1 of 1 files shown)


get-parity.sh
```diff
@@ -1,19 +1,61 @@
 #!/bin/bash
 # Copyright 2017 Parity Technologies (UK) Ltd.
+set -x
```
5chdn: should be removed after debugging?

get-parity.sh
```diff
@@ -1,19 +1,61 @@
...
 RELEASE="beta"
 ARCH=$(uname -m)
 VANITY_SERVICE_URL="https://vanity-service.parity.io/parity-binaries?architecture=$ARCH&format=markdown"
+DISTRO=$(lsb_release -c -s)
+
+check_alt() {
+    if [ "X${DISTRO}" == "X${2}" ]; then
+        echo
+        echo "## You seem to be using ${1} version ${DISTRO}."
+        echo "## This maps to ${3} \"${4}\"... Adjusting for you..."
```
5chdn: This is never reached on my Ubuntu 17.10

get-parity.sh
```diff
@@ -1,19 +1,61 @@
...
+		check_alt "Linux Mint"    "serena"   "Ubuntu" "xenial"
+		check_alt "Linux Mint"    "sonya"    "Ubuntu" "xenial"
+		check_alt "Linux Mint"    "sylvia"   "Ubuntu" "xenial"
+		check_alt "LMDE"          "betsy"    "Debian" "jessie"
+		check_alt "elementaryOS"  "luna"     "Ubuntu" "precise"
+		check_alt "elementaryOS"  "freya"    "Ubuntu" "trusty"
+		check_alt "elementaryOS"  "loki"     "Ubuntu" "xenial"
+		check_alt "Trisquel"      "toutatis" "Ubuntu" "precise"
+		check_alt "Trisquel"      "belenos"  "Ubuntu" "trusty"
+		check_alt "Trisquel"      "flidas"   "Ubuntu" "xenial"
```
5chdn: .

get-parity.sh
```diff
@@ -1,19 +1,61 @@
 #!/bin/bash
 # Copyright 2017 Parity Technologies (UK) Ltd.
+set -x
 RELEASE="beta"
 ARCH=$(uname -m)
 VANITY_SERVICE_URL="https://vanity-service.parity.io/parity-binaries?architecture=$ARCH&format=markdown"
+DISTRO=$(lsb_release -c -s)
+
+check_alt() {
+    if [ "X${DISTRO}" == "X${2}" ]; then
```
5chdn: Why the `X` prefix?

get-parity.sh
```diff
@@ -1,19 +1,61 @@
...
+		check_alt "Trisquel"      "flidas"   "Ubuntu" "xenial"
+		check_alt "BOSS"          "anokha"   "Debian" "wheezy"
+		check_alt "bunsenlabs"    "bunsen-hydrogen" "Debian" "jessie"
+		check_alt "Tanglu"        "chromodoris" "Debian" "jessie"
+		check_alt "Strech"        "strech" "Debian" "stretch"
+
+
+		if [ "$DISTRO" = "jessie" ] || [ "$DISTRO" = "wheezy" ] || [ "$DISTRO" = "stretch" ] ; then
+			PKG="debian"
+		elif [ "$DISTRO" = "xenial" ] || [ "$DISTRO" = "trusty" ]  || [ "$DISTRO" = "precise" ] ; then
```
5chdn: Missing `artful` and `bionic` (lts)

5chdn:
adding `artful` allowed me to upgrade on ubuntu, so LGTM apart from the minor comments above.

get-parity.sh
kirushik: @5chdn It's just an old-school best practive to prevent the value of `$DISTRO` to be interpreted as a parameter to `test` (for which `[` is just a syntactic sugar).
For example, if for some ungodly reason `$DISTRO` would be equal to string `"!"`, it might be interpreted as a negation for the upcoming conditional, resulting in an "incorrect syntax" error in runtime.

5chdn: Interesting!

PR #27, opened 2018-02-20
Filter sha256 returns from vanity service

ddorgan:


get-parity.sh
```diff
@@ -1,5 +1,6 @@
 #!/bin/bash
 # Copyright 2017 Parity Technologies (UK) Ltd.
+set -x
```
5chdn: Please remove.

LGTM - working on Ubuntu.

5chdn:
:+1:

PR #30, opened 2018-03-22
Silence curl download.

ddorgan:


get-parity.sh
```diff
@@ -108,7 +108,7 @@ install() {
 
     TMPDIR=$(mktemp -d)
     cd $TMPDIR
-	$(curl -Ss -O $DOWNLOAD_FILE)
+	$(curl -s -Ss -O $DOWNLOAD_FILE)
```
5chdn: `-s -Ss` 

really?

PR #34, opened 2018-06-26
Fixed extra pipe causing installs to fail

### LINKED ISSUE #32: Parity install script failing for specific version
The Parity client install script fails when using a specific version. Example:
bash <(curl https://get.parity.io -Lk) -r v1.10.6

dpkg-deb: error: '/tmp/tmp.VAwOMY2BP7/parity_1.10.6_ubuntu_amd64.deb.md5' is not a debian format archive

The problem is that the installation script will accidentally download the .md5 file instead of the .deb file.
https://raw.githubusercontent.com/paritytech/scripts/master/get-parity.sh

This began failing when the versioned vanity-service began returning .md5 files.
https://vanity-service.parity.io/parity-binaries?architecture=x86_64&format=markdown&os=linux&version=stable

Simple fix in installation script is to filter out md5 files:
MD=$(curl -Ss ${LOOKUP_URL} | grep -v sha256 | grep deb | grep -v md5)

codyborn:
Tested just this script locally by specifying version and default version

fix #32

### DIFF
--- a/get-parity.sh
+++ b/get-parity.sh
@@ -45,7 +45,7 @@ get_package() {
 	fi
 
 	if [ "$PKG" = "linux" -a "$LIBSSL" = "10" ]; then
-        MD=$(curl -Ss ${LOOKUP_URL} | grep -v sha256 | | grep -v md5 | grep deb)
+        MD=$(curl -Ss ${LOOKUP_URL} | grep -v sha256 | grep -v md5 | grep deb)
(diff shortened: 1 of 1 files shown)


5chdn:
That's weird. @fevo1971 can you look into this? Also, we don't have md5 anymore, maybe this needs some love in general?

5chdn:
cc @ddorgan

PR #42, opened 2018-09-18
add dependency gettext-base and remove aws from snapcraft image

gabreal:


docker-files-for-Gitlab-CI-rust/cross/snapcraft/Dockerfile
```diff
@@ -7,15 +7,23 @@ MAINTAINER Parity Technologies <devops@parity.io>
...
+        python \
+        zip \
+        curl \
+        expect \
+        gettext-base \
+        nodejs \
+        cmake \
+        wget && \
 	wget http://security.ubuntu.com/ubuntu/pool/universe/s/snapcraft/snapcraft_2.43.1_all.deb &&\
 	dpkg -i snapcraft_2.43.1_all.deb &&\
```
General-Beck: TODO - Until this version appears in the stable

gabreal: Yes, but we can leave it like this for now, can't we?

PR #50, opened 2018-12-07
disable substrate nightly build

gabreal:
substrate docker image and binaries are now build in the substrate repo with the dockerfile from there. Dependencies have changed that make the scripts job fail (clang). Therefore I suggest to schedule the job on the substrate repository if needed.


gabreal:
reworked

PR #56, opened 2018-12-30
getsubstrate - Update macOS and Ubuntu installation steps

ltfschoen:


bkchr:
In my opinion that is way too much for such a simple script. If you are a developer and does not have git configured, this script should not start doing this for you. Just exit and make the user configure it.
I'm also strongly against letting the script update software. Why should I want a script to run `apt upgrade` to update my whole distribution?

ltfschoen:
@bkchr Thanks for your feedback
* I agree that `apt upgrade` is overkill, but if we don't include `apt upgrade` shouldn't we at least install security updates for the user?
* I've removed the Git configuration steps. I initially added them in to save the user time, since I think that's all they need to do to start hacking on Substrate.
* If we exclude all the updating specific dependencies that they may need to use Substrate from the script, then if the user has older versions of software they they may need for hacking on Substrate (i.e. Yarn and Node.js) and the user wants to jump straight into hacking on a substrate front-end like https://github.com/polkadot-js/apps, then even though they thought they'd already "gotten Substrate" (and all the latest dependencies required for them to quickly get hacking) by running the one-stop-shop getsubstrate script, they'd discover in that repo's README that Yarn >=1.10.1 is required and that Node >=10.13.0 is recommended since some older versions won't work at all, so they'd lose time having to manually update those dependencies, when we could have done that for them.

ltfschoen:
@bkchr Is getsubstrate just meant to be a script for the Substrate Rust implementation?
If that's the case then I think we should edit the script and:
* **Keep** "Xcode Command Line Tools" step since "Homebrew" requires it.
* **Remove** "Node Version Manager (NVM)" and "Yarn", since they only need them if they wanted to hack on any of the Substrate/Polkadot front-ends 
* **Remove** "RBenv" step since macOS already has Ruby installed. And the only Substrate-related repo actually requires Ruby is the Polkadot.js webpage and requires Ruby since it's built using Jekyll https://github.com/polkadot-js/polkadot-js.github.io

What do you think?

get-substrate.sh
```diff
@@ -34,11 +36,121 @@ if [[ "$OSTYPE" == "linux-gnu" ]]; then
...
+	APP="Xcode Command Line Tools"
+	if xcode-select --install 2>&1 | grep "installed"; then
+		echo -e "Skipping, $APP already installed";
+	else
+		echo -e "Installing $APP ...";
+		xcode-select --install;
+	fi
+
+	APP="RBenv"
+	if ! rbenv 2>&1 | grep "command not found"; then
```
bkchr: If the user has a different locale, this `grep` will not work.
Please use something like: https://stackoverflow.com/questions/592620/how-to-check-if-a-program-exists-from-a-bash-script

get-substrate.sh
```diff
@@ -34,11 +36,121 @@ if [[ "$OSTYPE" == "linux-gnu" ]]; then
 elif [[ "$OSTYPE" == "darwin"* ]]; then
 	echo "Mac OS (Darwin) detected."
 
-	if ! which brew >/dev/null 2>&1; then
+	APP="Xcode Command Line Tools"
+	if xcode-select --install 2>&1 | grep "installed"; then
```
bkchr: Will probably also not work with a different locale.

get-substrate.sh
```diff
@@ -49,15 +161,50 @@ else
...
+	EXISTING_SUBSTRATE_VERSION=$(substrate --version)
+	echo -e "Substrate version $EXISTING_SUBSTRATE_VERSION detected on host machine"
 fi
-cargo install --force --git https://github.com/paritytech/substrate subkey
-cargo install --force --git https://github.com/paritytech/substrate substrate
 
+f=`mktemp -d`
+git clone https://github.com/paritytech/substrate-node-template $f
+# fetch new tags from remote repo
+git fetch --tags
```
bkchr: Which tags are you fetching here?

ltfschoen: I finally understand what you mean. I was meant to be fetching the tags of the substrate repo, not the substrate-node-template repo :)

ltfschoen:
@bkchr I've pushed changes to address your review comments. I've tested it on Ubuntu 16.04.2 and macOS Mojave.

PR #58, opened 2019-02-05
New Dockerfile

TriplEight:
- shrinked, 
- no useless pkges,
- pre-clone repos,
- new cache handler,
- less steps,
- ready for using cache in docker volumes

### DIFF
--- a/docker-files-for-Gitlab-CI-rust/parity-eth-linux-ci/Makefile
+++ b/docker-files-for-Gitlab-CI-rust/parity-eth-linux-ci/Makefile
@@ -1,0 +1,23 @@
+# IMAGE for build rustup
+# Created by Parity Technologies <devops-team@parity.io>
+USER=parity
+NAME=rust-linux
+TAG=stretch
+
+IMAGE=${USER}/${NAME}:${TAG}
+
+help:
+	@echo "Build rustup image."
+	@echo
+	@echo "Usage:"
+	@echo "  make [help]        - display this message"
+	@echo "  make image         - build rustup image"
+	@echo "  make push          - push build image to the regitry"
+
+image:
+	docker build --tag ${IMAGE} .
+
+push:
+	docker push ${IMAGE}
+
+.PHONY: help image push
--- a/docker-files-for-Gitlab-CI-rust/parity-eth-linux-ci/Makefile
+++ b/docker-files-for-Gitlab-CI-rust/parity-eth-linux-ci/Makefile
@@ -12,7 +12,7 @@ help:
 	@echo "Usage:"
 	@echo "  make [help]        - display this message"
 	@echo "  make image         - build rustup image"
-	@echo "  make push          - push build image to the regitry"
+	@echo "  make push          - push build image to the registry"
 
 image:
 	docker build --tag ${IMAGE} .
--- a/docker-files-for-Gitlab-CI-rust/parity-eth-linux-ci/README.md
+++ b/docker-files-for-Gitlab-CI-rust/parity-eth-linux-ci/README.md
@@ -1,0 +1,9 @@
+Docker image based on [RUST](https://www.rust-lang.org/) rust:stretch image (latest stable), [node.js](https://nodejs.org), for [GitLab CI runner](https://gitlab.com/gitlab-org/gitlab-ci-multi-runner).
+Usage:
+```
+linux-stable:
+    stage: build
+    image: parity/rust-linux:stretch
+    script:
+      - cargo build ...
+```
--- a/docker-files-for-Gitlab-CI-rust/rustup/Dockerfile
+++ b/docker-files-for-Gitlab-CI-rust/rustup/Dockerfile
@@ -1,46 +1,66 @@
-FROM rust:stretch
-LABEL maintainer="Parity Technologies <devops-team@parity.io>"
+FROM ubuntu:xenial
+MAINTAINER Parity Technologies <devops@parity.io>
+
+WORKDIR /build
 
 # install tools and dependencies
 RUN apt-get -y update && \
-	apt-get install -y --no-install-recommends \
-	software-properties-common curl git \
-	make cmake ca-certificates g++ rhash \
-	gcc pkg-config libudev-dev time
-
-# removed:
-# binutils binutils-dev snapcraft gettext file python build-essential zip dpkg-dev rpm libssl-dev openssl ruby-dev
+  apt-get install -y --no-install-recommends \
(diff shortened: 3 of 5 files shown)


docker-files-for-Gitlab-CI-rust/rustup/Dockerfile
```diff
@@ -1,66 +1,46 @@
...
 
-# show backtraces
-ENV RUST_BACKTRACE 1
+RUN cargo install cargo-audit sccache
 
 # cleanup
-RUN apt autoremove -y
-RUN apt clean -y
-RUN rm -rf /tmp/* /var/tmp/*
+RUN echo cleanup && \
```
kirushik: @TriplEight This actually makes little sense, since you'll still have to download a Docker layer with all those removed files — and then another one, to erase them.

If those cache files pollute our builds, we should try to stick the cleaning bash expressions in the same `RUN` command which creates them.
If we don't care — let's just keep them be, and our Docker images will be even slightly smaller size and faster to download and access (because of less layers).

docker-files-for-Gitlab-CI-rust/rustup/Dockerfile
```diff
@@ -1,66 +1,46 @@
...
+	rm -rf /tmp/* /var/tmp/*
 
-#compiler ENV
+# compiler ENV
 ENV CC gcc
 ENV CXX g++
+ENV CARGO_TARGET x86_64-unknown-linux-gnu
+ENV CARGO_HOME /parity-ethereum/.cargo/
+# cache handler
+ENV RUSTC_WRAPPER sccache
```
kirushik: Huge cheers for `sccache`!
:clap:

docker-files-for-Gitlab-CI-rust/rustup/Dockerfile
```diff
@@ -1,66 +1,46 @@
...
+# show backtraces
+ENV RUST_BACKTRACE 1
+
+# pre-download repositories
+RUN git clone https://github.com/paritytech/parity-ethereum.git . && \
```
kirushik: Does this bring any noticeable speedup actually? How much?
I probably understand the rationale, but this just rubs me the wrong way.

fevo1971:
what happend to the aws-tools needed to for releases?

docker-files-for-Gitlab-CI-rust/rustup/Dockerfile
TriplEight: Yes, but I feel like we will face some issues with it in future.

docker-files-for-Gitlab-CI-rust/rustup/Dockerfile
TriplEight: in CI config there are 
```
  GIT_STRATEGY:                    fetch
  GIT_SUBMODULE_STRATEGY:          recursive
```
settings now, it will help to economy ~ a minute on each pipeline, submodules really take time to clone.

TriplEight:
> what happend to the aws-tools needed to for releases?

good question, but unfortunately I didn't have a chance to test release while not on production branch, think I'll need some help with it.

docker-files-for-Gitlab-CI-rust/rustup/Dockerfile
TriplEight: makes sense, will remove this cleanup on the next commit.

gabreal:
> what happend to the aws-tools needed to for releases?

I created a [aws cli container](https://hub.docker.com/r/parity/awscli) and suggest to use it for publishing purposes like [this](https://github.com/paritytech/substrate/blob/master/.gitlab-ci.yml#L187)

docker-files-for-Gitlab-CI-rust/rustup/Dockerfile
kirushik: @gabreal you can point it to, say S3 or Google Cloud Storage; no dedicated server is needed when working in this mode.

docker-files-for-Gitlab-CI-rust/rustup/Dockerfile
```diff
@@ -1,66 +1,46 @@
...
+		git submodule update --init --recursive
 
-#snapcraft ENV
-ENV LC_ALL=C.UTF-8
-ENV LANG=C.UTF-8
+VOLUME /parity-ethereum/target $CARGO_HOME
```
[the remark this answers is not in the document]
TriplEight: I had a plan of a huge unified container, but it doesn't seem real yet.

docker-files-for-Gitlab-CI-rust/parity-eth-linux-ci/Makefile
```diff
@@ -0,0 +1,23 @@
...
+
+IMAGE=${USER}/${NAME}:${TAG}
+
+help:
+	@echo "Build rustup image."
+	@echo
+	@echo "Usage:"
+	@echo "  make [help]        - display this message"
+	@echo "  make image         - build rustup image"
+	@echo "  make push          - push build image to the regitry"
```
ltfschoen: nickpick typo 'registry'

TriplEight: thanks!

docker-files-for-Gitlab-CI-rust/parity-eth-linux-ci/README.md
```diff
@@ -0,0 +1,9 @@
+Docker image based on [RUST](https://www.rust-lang.org/) rust:stretch image (latest stable), [node.js](https://nodejs.org), for [GitLab CI runner](https://gitlab.com/gitlab-org/gitlab-ci-multi-runner).
+Usage:
+```
+linux-stable:
+    stage: build
+    image: parity/rust-linux:stretch
```
gabreal: just as a fincal comment (and too late):
as the dockerfile is tailored for parity-ethereum, probably the docker tag or image string and the documentation should reflect that @TriplEight

TriplEight: you mean to change the maintainer, makefile maintainer, that would help someone to fix if something.
But placing a nickname in the container name, it would be at least not informative. I agree that came up with not the best name for the container, maybe you have a better one?

gabreal: well I mean if the containers purpose is entirely to build parity-ethereum, why not call the image rust-parity-ethereum or tag the rust-linux image stretch-parity-ethereum

PR #62, opened 2019-03-20
add appropriate pacman flags in get-substrate.sh for arch linux

lovelaced:
The --noconfirm flag is necessary for this script to work if one curls it or otherwise uses it in a noninteractive manner. I believe this to be Arch's correct way of doing this.

### DIFF
--- a/get-substrate.sh
+++ b/get-substrate.sh
@@ -15,7 +15,7 @@ if [[ "$OSTYPE" == "linux-gnu" ]]; then
 		echo "Please refer to https://github.com/paritytech/substrate for setup information."
 	elif [ -f /etc/arch-release ]; then
 		echo "Arch Linux detected."
-		$MAKE_ME_ROOT pacman -Sy cmake gcc openssl-1.0 pkgconf git clang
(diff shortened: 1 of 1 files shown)


get-substrate.sh
```diff
@@ -15,7 +15,7 @@ if [[ "$OSTYPE" == "linux-gnu" ]]; then
 		echo "Please refer to https://github.com/paritytech/substrate for setup information."
 	elif [ -f /etc/arch-release ]; then
 		echo "Arch Linux detected."
-		$MAKE_ME_ROOT pacman -Sy cmake gcc openssl-1.0 pkgconf git clang
+		$MAKE_ME_ROOT pacman -Sy --noconfirm cmake gcc openssl-1.0 pkgconf git clang
```
DemiMarie-parity:
```suggestion
		$MAKE_ME_ROOT pacman -Syu --noconfirm cmake gcc openssl-1.0 pkgconf git clang
```

Arch [does not support partial upgrades](//wiki.archlinux.org/index.php/System_maintenance#Partial_upgrades_are_unsupported), so `pacman -Syu` **must** be run before installing any packages.

lovelaced:
This was fixed in #78  but was never closed.

PR #66, opened 2019-03-28
Android dockerfile update

TriplEight:
- smaller
- optimized
- with cache


tomaka:
The indentation is inconsistent and there are commented out lines of code, but looks good other than that.

TriplEight:
> The indentation is inconsistent and there are commented out lines of code, but looks good other than that.

Indentation shows related commands like installations and ripped long commands.
Commented code is to understand where it's needed, to navigate in this wild.
Looks better than before, but this build environment is insane, would be very happy to make it simpler.

tomaka:
> Indentation shows related commands like installations and ripped long commands.

What I mean is you're mixing spaces and tabs even at the first level of indentation.

TriplEight:
oh noes, will fix it right away.

PR #70, opened 2019-04-11
get-substrate.sh should set default toolchain when installing rust

gabreal:


get-substrate.sh
```diff
@@ -52,6 +52,7 @@ fi
 if ! which rustup >/dev/null 2>&1; then
 	curl https://sh.rustup.rs -sSf | sh -s -- -y
 	source ~/.cargo/env
+  rustup default stable
```
TriplEight: Indent please

get-substrate.sh
```diff
@@ -52,6 +52,7 @@ fi
 if ! which rustup >/dev/null 2>&1; then
 	curl https://sh.rustup.rs -sSf | sh -s -- -y
 	source ~/.cargo/env
+	srustup default stable
```
TriplEight: typo ^^

bkchr:
Hmm, I really don't like to force people into what we think is correct. If we need stable, then we need to run `cargo +stable build`, but otherwise I would like to not alter systems by users.

bkchr:
Okay, it is just for installing. Forget what I said.

PR #86, opened 2019-05-17
Parity Builder restyling

TriplEight:
- lost some variables which should be in CI config
- lost not needed `wget` and `software-properties-common`
- name
- dir to match name
- labels
- README
- scripts CI


TriplEight:
I've tested everything locally. Also, the previous tag will stay there for a while.

PR #91, opened 2019-08-12
add to sccache features redis

General-Beck:
All the images which are used in parity-ethereum build process are now being published to our Gitilab docker container registry.
These images now are making use of `sccache` with Redis to avoid the data race condition when pruning old cache pieces.
Metadata was also edited to show the current addresses of the images.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -62,7 +62,7 @@ variables:
   <<:                              *docker_build
   before_script:
     - test -z "$DOCKERIMAGE" && echo "DOCKERIMAGE must be defined" && exit 1
-    - docker login -u $CI_DEPLOY_USER -p $CI_DEPLOY_PASSWORD $CI_REGISTRY
+    - docker login -u $CI_DEPLOY_USER -p $CI_DEPLOY_PASSWORD $CI_REGISTRY/pub/ci
     - docker info
   script:
     - cd dockerfiles/$DOCKERFILE_DIR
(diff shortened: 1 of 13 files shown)


.gitlab-ci.yml
```diff
@@ -58,6 +58,26 @@ variables:
   environment:
     name: parity-build
 
+.build_to_own_reg:                 &docker_build_to_own_reg
+  <<:                              *docker_build
+  before_script:
+    - test -z "$DOCKERIMAGE" && echo "DOCKERIMAGE must be defined" && exit 1
+    - docker login -u $CI_REGISTRY_USER -p $CI_REGISTRY_PASSWORD $CI_REGISTRY
```
gabreal: would be cool to avoid this kind of login somehow

.gitlab-ci.yml
```diff
@@ -58,6 +58,26 @@ variables:
...
+  script:
+    # $CI_REGISTRY_IMAGE == registry.parity.io/parity/infrastructure/scripts/
+    # $CI_REGISTRY == registry.parity.io/
```
gabreal: where are these variables defined then?

.gitlab-ci.yml
TriplEight: these are built-in Gitlab variables that change every job actually, AFAIU it's the most secure way to login to gitlab registry in order to `push` into it.

.gitlab-ci.yml
TriplEight: two lines below :)

TriplEight: I left those comments there just for the information.

.gitlab-ci.yml
gabreal: nice :+1:

.gitlab-ci.yml
gabreal: but don't they have to be defined somewhere - the url registry.parity.io could also be different.

TriplEight: nope, `CI_REGISTRY` is the gitlab.parity.io wide and equals registry.parity.io 
and `CI_REGISTRY_IMAGE` will always be `CI_REGISTRY` + current project url

TriplEight: so `registry.parity.io` is defined as root registry in gitlab's `/etc/gitlab/gitlab.rb`

PR #101, opened 2019-10-04
Remove need for `init.sh` script in `get-substrate.sh`

shawntabrizi:
This changes the `get-substrate.sh` script to automatically install add the `wasm32` target to rust nightly, and installs `wasm-gc`.

This means we don't need to call `init.sh` anymore for new users as this already does those steps for them.

### DIFF
--- a/get-substrate.sh
+++ b/get-substrate.sh
@@ -67,19 +67,19 @@ else
 	rustup default stable
 fi
 
+rustup target add wasm32-unknown-unknown --toolchain nightly
+# Install wasm-gc. It's useful for stripping slimming down wasm binaries.
+command -v wasm-gc || \
+	cargo +nightly install --git https://github.com/alexcrichton/wasm-gc --force
+
 if [[ "$1" == "--fast" ]]; then
 	echo "Skipped cargo install of 'substrate' and 'subkey'"
 	echo "You can install manually by cloning the https://github.com/paritytech/substrate repo,"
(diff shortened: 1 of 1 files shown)


get-substrate.sh
```diff
@@ -67,19 +67,19 @@ else
 	rustup default stable
 fi
```
riusricardo:
```suggestion
rustup update nightly
```

riusricardo: It needs to get nightly if it is not installed.

get-substrate.sh
```diff
@@ -67,19 +67,19 @@ else
...
 
+rustup target add wasm32-unknown-unknown --toolchain nightly
```
riusricardo:
```suggestion
rustup update nightly
rustup target add wasm32-unknown-unknown --toolchain nightly
```

riusricardo: Needs to get nightly if it is not installed

shawntabrizi: I tested it on a "new" computer and `rustup update nightly` was enough to install nightly if it wasnt there already.

riusricardo: Yes you are right, that is why I suggested it here. You missed the `rustup update nightly` command. 
Doesn't matter anymore, I committed that fix to the script.

shawntabrizi: oh, nice, thank you :)

PR #105, opened 2019-11-07
switch from stretch to buster. add latest gcc, g++, clang, cmake 

General-Beck:
cmake 3.16.0
For non `parity-ethereum` CI images:
stretch -> buster 
clang - 9 
gcc-8.3
g++-8.3


dockerfiles/parity-ci-linux/Dockerfile
```diff
@@ -32,8 +32,13 @@ RUN set -eux; \
...
 # libssl-dev is needed for cargo-audit and sccache
 		g++ libssl-dev gcc libc6-dev \
-		make cmake libudev-dev ca-certificates \
-		git pkg-config curl time rhash; \
+		make libudev-dev ca-certificates \
+		git pkg-config curl time rhash software-properties-common apt-transport-https; \
+	apt-add-repository ppa:ubuntu-toolchain-r/test; \
+	apt-add-repository 'deb https://apt.kitware.com/ubuntu/ xenial main'; \
+	apt-get -y update; \
+	apt-get install -y  --allow-unauthenticated gcc-9 g++-9 cmake; \
```
DemiMarie-parity: noooooo!!!!!! `--allow-unauthenticated` is a remote root exploit by design!

More seriously, if we are going to do this, we need to be absolutely certain that this image never gets its hands on anything sensitive (either regarding confidentiality or integrity).  Could we upgrade to a newer version of the OS instead?

DemiMarie-parity: Another option is to install GCC from source, since the GNU project provides signed release tarballs.

gabreal: gcc-9 is available in debian [testing](https://packages.debian.org/bullseye/gcc-9). I suggest to use it from there instead from a unknown or potentially unmaintained third party.

gabreal: didn't see that it's ubuntu based, so instead I suggest the [disco](https://packages.ubuntu.com/disco/gcc-9) package.

TriplEight:
@DemiMarie-parity please tell me why do you want to update gcc and g++ here? I'm not sure we will guarantee Ubuntu 14.04 support for the client. I already had this discussion when I tried to change the base image to Debian stretch.

DemiMarie-parity:
@TriplEight because old versions of g++ don’t support modern C++ features.

dockerfiles/parity-ci-linux/Dockerfile
General-Beck: please see the latest changes

dockerfiles/base-ci-linux/Dockerfile
```diff
@@ -22,6 +22,9 @@ WORKDIR /builds
 
 # install tools and dependencies
 COPY clang9.key /etc/apt/trusted.gpg.d/debian-archive-jessie-automatic.gpg
+#download and install cmake
+ADD "https://github.com/Kitware/CMake/releases/download/v3.16.0/cmake-3.16.0-Linux-x86_64.sh" cmake.sh
```
gabreal: why not use https://packages.debian.org/buster/cmake ? it's version 3.13, is 3.19 really necessary, I think it will be easy to forget to update patch versions like that.

dockerfiles/base-ci-linux/Dockerfile
```diff
@@ -22,6 +22,9 @@ WORKDIR /builds
 
 # install tools and dependencies
 COPY clang9.key /etc/apt/trusted.gpg.d/debian-archive-jessie-automatic.gpg
```
gabreal: I didn't see that before, but is that key for this http://apt.llvm.org/stretch/ repo from below. the name is misleading imo.
btw there would also be a debian package in the testing suite i.e. https://packages.debian.org/bullseye/clang-9

dockerfiles/base-ci-linux/Dockerfile
General-Beck: Our Base Image is built on the stretch distribution. And there the maximum version is 3.9. C ++ 17 support has been included since version 3.10. Well, in this case, we take the version from the upstream, so it seems to me that this option is preferable

dockerfiles/base-ci-linux/Dockerfile
General-Beck: The name of the key was suggested by @TriplEight  and I quite understand.
Questions, with the increase of the version from `stretch` to `buster` or from `xenial` to `bionic` I already raised in the channel devops.

TriplEight:
@gabreal @General-Beck I'd better switch the base to `buster` and follow it with the newer versions.

General-Beck:
I'll make a new pull request for `buster`

dockerfiles/base-ci-linux/Dockerfile
DemiMarie-parity: @General-Beck I agree with @gabreal here, especially if a backport is available in Debian.

dockerfiles/parity-ci-linux/README.md
```diff
@@ -1,6 +1,6 @@
 # Parity-ci-linux
 
-Docker image based on [Ubuntu:xenial](https://hub.docker.com/_/ubuntu).
```
TriplEight: Let's return parity-eth CI to xenial unless we are sure the compiled binary will work on xenial.

dockerfiles/base-ci-linux/Dockerfile
TriplEight: @gabreal , what name do you suggest? I thought it's pretty much straightforward, we install `clang-9` and `libclang-9-dev` from that repo.

dockerfiles/base-ci-linux/Dockerfile
TriplEight: Now, as we are on `buster`, we shouldn't install `cmake`, we have the right version built-in, right?

DemiMarie-parity: @TriplEight you are correct.

dockerfiles/base-ci-linux/README.md
```diff
@@ -10,23 +10,23 @@ Used to build and test Substrate-based projects.
...
-- `g++`
-- `gcc`
-- `git`
-- `libssl-dev`
-- `make`
-- `pkg-config`
-- `rhash`
-- `rust-builder`
-- `time`
+-   `clang-9`
```
DemiMarie-parity: Why the extra space?

dockerfiles/ink-ci-linux/README.md
```diff
@@ -8,37 +8,37 @@ Used to build and test ink!.
...
-- `g++`
-- `gcc`
-- `git`
-- `libssl-dev`
-- `make`
-- `pkg-config`
-- `rhash`
-- `rust-builder`
-- `time`
+-   `clang-8`
```
DemiMarie-parity: Why the extra whitespace?

dockerfiles/parity-ci-android/Dockerfile
```diff
@@ -17,6 +17,9 @@ dockerfiles/parity-ci-android/README.md" \
 	io.parity.image.revision="${VCS_REF}" \
 	io.parity.image.created="${BUILD_DATE}"
 
+# download and install cmake
+ADD "https://github.com/Kitware/CMake/releases/download/v3.16.0/cmake-3.16.0-Linux-x86_64.sh" cmake.sh
```
DemiMarie-parity: We don’t need this as CMake 3.13 is enough for now.

dockerfiles/parity-ci-linux/Dockerfile
```diff
@@ -26,6 +26,10 @@ ENV RUSTUP_HOME=/usr/local/rustup \
 
 # download rustup
 ADD "https://static.rust-lang.org/rustup/dist/x86_64-unknown-linux-gnu/rustup-init" rustup-init
+# download and install cmake
+ADD "https://github.com/Kitware/CMake/releases/download/v3.16.0/cmake-3.16.0-Linux-x86_64.sh" cmake.sh
```
DemiMarie-parity: ditto

dockerfiles/rust-builder/Dockerfile
```diff
@@ -25,14 +25,17 @@ COPY clang9.key /etc/apt/trusted.gpg.d/debian-archive-jessie-automatic.gpg
 COPY config /root/.cargo/config
 # gcc and clang are necessary to be able to compile crates that have build scripts
 # Most notably, clang 9 has WASM support enabled by default.
+# download and install cmake
+ADD "https://github.com/Kitware/CMake/releases/download/v3.16.0/cmake-3.16.0-Linux-x86_64.sh" cmake.sh
```
DemiMarie-parity: ditto

dockerfiles/base-ci-linux/Dockerfile
General-Beck: I thought we would use the upstream version of cmake. In CI then we will have a difference in versions on xenial, windows and osx.
xenial - 3.5.1
buster - 3.13.4
windows - 3.16
osx - 3.15
I think this option is the best in our case.

dockerfiles/base-ci-linux/README.md
General-Beck: linter in Atom

General-Beck:
@DemiMarie-parity @gabreal @TriplEight So my last question, we will be using cmake version 3.16. everywhere or will we have a different in versions?

dockerfiles/parity-ci-linux/README.md
General-Beck: I agree, for parity-ethereum, we will leave this version for now.

TriplEight:
I vote for one version everywhere. Seems like there's no other good way but installing it everywhere (xenial, buster, win, mac) from the upstream. @gabreal , @DemiMarie-parity any objections?

DemiMarie-parity:
@TriplEight Not if Kitware provides signed releases and we verify them.  I just don’t want to rely only on the TLS to the server.

General-Beck:
@DemiMarie-parity [3.16.0 - verified](https://github.com/Kitware/CMake/releases/tag/v3.16.0)

General-Beck:
[Parity-ethereum - passed](https://gitlab.parity.io/General-Beck/parity-ethereum/pipelines/71178) binary tested (only cmake updated)

DemiMarie-parity:
@General-Beck that works!

General-Beck:
waiting for
[Polkadot (buster+make)](https://gitlab.parity.io/General-Beck/polkadot/pipelines/71175)
[Substrate (buster+cmake)](https://gitlab.parity.io/General-Beck/substrate/pipelines/71174)

dockerfiles/base-ci-linux/Dockerfile
```diff
@@ -20,15 +20,14 @@ dockerfiles/rust-builder/README.md" \
...
-#download and install cmake
-ADD "https://github.com/Kitware/CMake/releases/download/v3.16.0/cmake-3.16.0-Linux-x86_64.sh" cmake.sh
-RUN sh cmake.sh --skip-license --prefix=/usr/ --exclude-subdir&&rm cmake.sh&&cmake --version
 # config for wasm32-unknown-unknown
 COPY config /root/.cargo/config
 # gcc and clang are necessary to be able to compile crates that have build scripts
 # Most notably, clang 9 has WASM support enabled by default.
+COPY clang9.key /etc/apt/trusted.gpg.d/debian-archive-jessie-automatic.gpg
+# download cmake 
+ADD "https://github.com/Kitware/CMake/releases/download/v3.16.0/cmake-3.16.0-Linux-x86_64.sh" cmake.sh
```
DemiMarie-parity: We should use the signed release file instead.

dockerfiles/parity-ci-android/Dockerfile
```diff
@@ -17,9 +17,8 @@ dockerfiles/parity-ci-android/README.md" \
 	io.parity.image.revision="${VCS_REF}" \
 	io.parity.image.created="${BUILD_DATE}"
 
-# download and install cmake
+# download cmake
 ADD "https://github.com/Kitware/CMake/releases/download/v3.16.0/cmake-3.16.0-Linux-x86_64.sh" cmake.sh
```
DemiMarie-parity: Again, we need to use the signed release file.

dockerfiles/parity-ci-linux/Dockerfile
```diff
@@ -26,12 +26,10 @@ ENV RUSTUP_HOME=/usr/local/rustup \
 
 # download rustup
 ADD "https://static.rust-lang.org/rustup/dist/x86_64-unknown-linux-gnu/rustup-init" rustup-init
-# download and install cmake
+# download cmake 
 ADD "https://github.com/Kitware/CMake/releases/download/v3.16.0/cmake-3.16.0-Linux-x86_64.sh" cmake.sh
```
DemiMarie-parity: Same

dockerfiles/rust-builder/Dockerfile
```diff
@@ -19,15 +19,14 @@ dockerfiles/rust-builder/README.md" \
...
-# install tools and dependencies
-COPY clang9.key /etc/apt/trusted.gpg.d/debian-archive-jessie-automatic.gpg
 # config for wasm32-unknown-unknown
 COPY config /root/.cargo/config
 # gcc and clang are necessary to be able to compile crates that have build scripts
 # Most notably, clang 9 has WASM support enabled by default.
-# download and install cmake
+COPY clang9.key /etc/apt/trusted.gpg.d/debian-archive-jessie-automatic.gpg
+# download cmake 
 ADD "https://github.com/Kitware/CMake/releases/download/v3.16.0/cmake-3.16.0-Linux-x86_64.sh" cmake.sh
```
DemiMarie-parity: Same

dockerfiles/base-ci-linux/Dockerfile
gabreal: The name is misleading as the resulting filename `debian-archive-jessie-automatic.gpg` gives no clue that it is actually the key for the clang repo.

gabreal: Now, that you switched the base image to `buster`, I don't see why it would be needed to add llvms repository at all. IMO it's better to use the distro packages because they are supported by the distros' community and have been audited by a third-parity which is not the software vendor. But it's not so important and I can live with it if the key is properly verified, it's just another point where things can go wrong.

dockerfiles/base-ci-linux/Dockerfile
gabreal: I'd say `keep it simple`. Use the distro packages everywhere from the package managers we use (`apt`, `brew`, `chocolate`). It should be possible to agree on a minimum version and I assume versions should be downwards compatible. It's also giving me more confidence than running arbitrary shell scripts for creating our build environment. How can it be guaranteed that it will work well together with possible future updates of other packages? Who will be notified if there is a new upstream version with potential bug fixes?

dockerfiles/rust-builder/Dockerfile
```diff
@@ -1,5 +1,5 @@
 # DockerHub's official https://hub.docker.com/_/rust
-FROM rust:slim-stretch
+FROM rust:slim-buster
```
DemiMarie-parity: We should check that this image is signed by Docker.

TriplEight: Can we please do this in any other PR? It should be done in CI anyways.

General-Beck: https://github.com/paritytech/devops/issues/386

General-Beck:
parity-ethereum (check, test's, build x86_64&android) -  [passed](https://gitlab.parity.io/General-Beck/parity-ethereum/pipelines/71178)
parity-ethereum (build arm64,armhf,i386,windows,macos) [passed](https://gitlab.parity.io/General-Beck/parity-ethereum/pipelines/71759)
polkadot (check, test, build) - [passed](https://gitlab.parity.io/General-Beck/polkadot/pipelines/71175)
substarte (check, test, build) - [waiting](https://gitlab.parity.io/General-Beck/substrate/pipelines/71762)

dockerfiles/parity-ci-linux/README.md
```diff
@@ -1,6 +1,6 @@
...
+Docker image based on [Ubuntu:xenial](https://hub.docker.com/_/debian).
```
TriplEight:
```suggestion
Docker image based on [Ubuntu:xenial](https://hub.docker.com/_/ubuntu).
```

dockerfiles/utility/README.md
```diff
@@ -0,0 +1,7 @@
+clang9.key - key for install clang9 from [https://apt.llvm.org](https://apt.llvm.org)
```
TriplEight:
```suggestion
`debian-llvm-clang9.key` - debian repo key to install `clang-9` and `libclang-9-dev` from [https://apt.llvm.org](https://apt.llvm.org)
```

dockerfiles/utility/README.md
```diff
@@ -0,0 +1,7 @@
...
+
+cmake.sh - stable version cmake from [https://github.com/Kitware/CMake/releases](https://github.com/Kitware/CMake/releases)
```
TriplEight:
```suggestion
`cmake.sh` - stable `cmake` from [https://github.com/Kitware/CMake/releases](https://github.com/Kitware/CMake/releases)
```

dockerfiles/README.md
```diff
@@ -1,4 +1,26 @@
-Rust 1.36.0
+Rust 1.39.0
```
TriplEight:
```suggestion
Most of these Dockerfiles rely on `utility` dir to `COPY` some files from it.
This means if you want to build these Dockerfiles you **must** do it from `dockerfiles` dir.
Like that:

cd dockerfiles
docker build --build-arg VCS_REF="12345" --build-arg BUILD_DATE="(date -u '+%Y-%m-%dT%H:%M:%SZ')" --tag grcov .
docker build --no-cache
        --build-arg VCS_REF="12345"
        --build-arg BUILD_DATE="$(date +%Y%m%d)"
        --tag my-local:substrate
        --file <DOCKERFILE_DIR>/Dockerfile .

Rust 1.39.0
```

.gitlab-ci.yml
```diff
@@ -42,15 +42,16 @@ variables:
     - docker login -u "${Docker_Hub_User_Parity}" -p "${Docker_Hub_Pass_Parity}"
     - docker info
   script:
-    - cd dockerfiles/$DOCKERFILE_DIR
     # - docker pull $CONTAINER_IMAGE:$CONTAINER_TAG || true
     # - docker build --cache-from $CONTAINER_IMAGE:$CONTAINER_TAG --tag $CONTAINER_IMAGE:$CI_BUILD_REF --tag $CONTAINER_IMAGE:$CONTAINER_TAG .
+    - cd dockerfiles/
```
gabreal: dosn't that copy the whole directory into the build context of docker build?

gabreal: I see - the `utitlity` directory, just ignore my question.

dockerfiles/base-ci-linux/Dockerfile
```diff
@@ -20,24 +20,32 @@ dockerfiles/rust-builder/README.md" \
...
 		> /etc/apt/sources.list.d/llvm.list; \
-	echo "deb-src http://apt.llvm.org/stretch/ llvm-toolchain-stretch-9 main" \
+	echo "deb-src http://apt.llvm.org/buster/ llvm-toolchain-buster-9 main" \
 		>> /etc/apt/sources.list.d/llvm.list; \
-	DEBIAN_FRONTEND=noninteractive apt-get -y update; \
-	DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends \
-		g++ libssl-dev gcc clang-9 cmake make \
+	apt-get -y update; \
+	apt-get install -y --no-install-recommends \
+		g++ libssl-dev gcc clang-9 libclang-9-dev make \
```
gabreal: why do the whole llvm repo thing here when pulling in clang from the distro at line no 12?

dockerfiles/base-ci-linux/Dockerfile
```diff
@@ -20,24 +20,32 @@ dockerfiles/rust-builder/README.md" \
...
 # config for wasm32-unknown-unknown
-COPY config /root/.cargo/config
+COPY utility/base-ci-linux-config /root/.cargo/config
+
 # gcc and clang are necessary to be able to compile crates that have build scripts
 # Most notably, clang 9 has WASM support enabled by default.
+COPY utility/debian-llvm-clang9.key /etc/apt/trusted.gpg.d/debian-archive-llvm.gpg
+
+# copy cmake
+COPY utility/cmake.sh cmake.sh
```
gabreal: that's a 38Mb binary-containing script in the repo?

dockerfiles/base-ci-linux/Dockerfile
```diff
@@ -20,24 +20,32 @@ dockerfiles/rust-builder/README.md" \
...
-	DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends \
-		g++ libssl-dev gcc clang-9 cmake make \
+	apt-get -y update; \
+	apt-get install -y --no-install-recommends \
+		g++ libssl-dev gcc clang-9 libclang-9-dev make \
 		git pkg-config curl time rhash; \
 # set a link to clang-9
 	ln -s /usr/bin/clang-9 /usr/bin/clang; \
+# install cmake
+	./cmake.sh --skip-license --prefix=/usr/ --exclude-subdir; \
```
gabreal: IMO it makes no sense to rebuild the same archive-containing script again every night. The result won't change.

dockerfiles/base-ci-linux/Dockerfile
TriplEight: line 12 is the description of the container.

dockerfiles/base-ci-linux/Dockerfile
TriplEight: yep, looks disgusting

dockerfiles/base-ci-linux/Dockerfile
TriplEight: what do you propose here?

dockerfiles/base-ci-linux/Dockerfile
kirushik: But is it worse than downloading the said script over the web every time (both buildtime-wise and security-wise)?

dockerfiles/base-ci-linux/Dockerfile
gabreal: use a precompiled debian package like https://packages.debian.org/bullseye/cmake
cmake patch release `3.16.2` is out and it will not be included here (instead the same outdated version is build over and over again). If it's really a big concern about using the same version everywhere, why not use this one - makes things way easier.

dockerfiles/base-ci-linux/Dockerfile
gabreal: the file indicates in no way which version it contains. one has to open a 38Mb file and look inside (or run it), that makes it hard to track which version it is and updating potenital security fixes will most likely not happen. attaching to some kind of upstream update mechanism at least for the same minor release would be better imo.

gabreal: and for the binary blob - that is way too big for a source repository. just imagine if one would update the file more than once - the repository will grow huge. a better approach would be to use a prebuild version uploaded into say a s3 bucket (or even use a gitlab ci for that) and revert the whole repo to get it back small and handy again.

dockerfiles/base-ci-linux/Dockerfile
TriplEight: that definitely makes sense, mill mark myself to do that later. Thanks!

PR #127, opened 2019-12-11
docker container build with redis_exporter

gabreal:
deployment suggestion: https://mehmandarov.com/start-docker-containers-automatically/ running it inside docker with systemd.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -250,10 +250,15 @@ kubetools:
 
 
 
-
-
-
-
+redis_exporter:
+  <<: *docker_build
+  variables:
+    CONTAINER_IMAGE:               $CI_JOB_NAME
+    CONTAINER_TAG:                 latest
+    DOCKERFILE_DIR:                $CI_JOB_NAME
+  only:
+    variables:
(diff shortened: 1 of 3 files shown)


lovelaced:
Thanks for this! Building in a docker image seems pretty easy and convenient. Out of curiosity, why not just rely on the docker daemon to autostart the image: https://docs.docker.com/config/containers/start-containers-automatically/
I found some gotchas about using systemd to control running docker containers in this post, but it is a bit "old" in terms of docker maturity: https://blog.container-solutions.com/running-docker-containers-with-systemd

gabreal:
@lovelaced  thanks for the links. My initial thoughts were that systemd would be more transparent and mature in handling system level jobs. Also that you would not have a dependency to think of anything when restarting the docker daemon. But another good argument from the article above:

> Using systemd or an equivalent is particularly useful if you have another, possibly non-containerized service that is dependent on the container.

But I guess I would prefer to run it via systemd even when running containerized redis, or would it be possible to declare dependencies in docker?

TriplEight:
:+1:

gabreal:
That's the deployment: https://gitlab.parity.io/parity/devops/merge_requests/45

PR #130, opened 2019-12-17
container scanner

TriplEight:
introduce a built-in container scanner

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -81,6 +81,11 @@ variables:
     - docker push $CI_REGISTRY_IMAGE/$CONTAINER_IMAGE:$CONTAINER_TAG
 
 
+# triggers the container scanning built-in tool
+include:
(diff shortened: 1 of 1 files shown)


gabreal:
interesting tool - I have no clue how it works but it seems to trigger a ci pipeline (which fails) for pull requests now:
https://gitlab.parity.io/parity/infrastructure/scripts/-/jobs/334039
https://gitlab.parity.io/parity/infrastructure/scripts/pipelines/72120

how does it work? is it maybe worth it to add something to the repo readme?

TriplEight:
It connects a built-in part of yaml, which describes a service, it does everything internally and automatically, yet vaguely. I couldn't make it work from the first attempt, so reverted it in https://github.com/paritytech/scripts/pull/131 .
I think it'd be better switch these containers to be built by `buildah` and check them with it's built-in checker.

PR #142, opened 2020-01-28
lint for get scripts

TriplEight:
Just linted these scrips and fixed one typo.

### DIFF
--- a/get-substrate.sh
+++ b/get-substrate.sh
@@ -1,6 +1,9 @@
+#!/bin/bash
+# Copyright 2015-2020 Parity Technologies (UK) Ltd.
+
 if [[ "$OSTYPE" == "linux-gnu" ]]; then
 	set -e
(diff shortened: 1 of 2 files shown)


get-substrate.sh
```diff
@@ -75,17 +78,17 @@ if [[ "$1" == "--fast" ]]; then
...
 fi
 
-f=`mktemp -d`
-git clone https://github.com/paritytech/substrate-up $f
-cp -a $f/substrate-* ~/.cargo/bin
-cp -a $f/polkadot-* ~/.cargo/bin
+f=$(mktemp -d)
+git clone https://github.com/paritytech/substrate-up "$f"
+cp -a "$f"/substrate-* ~/.cargo/bin
+cp -a "$f"/polkadot-* ~/.cargo/bin
```
riusricardo: We should delete this lines. This script are deprecated and aren't used in v2.0

PR #151, opened 2020-02-12
Update ARM lld

riusricardo:
Updates rust llvm branch.
Deletes duplicated file:
~~/install-wasm-binaries.sh~~ -> /wasm-utils/install-wasm-binaries.sh


wasm-utils/install-rust-lld-ARM.sh
```diff
@@ -1,99 +1,96 @@
...
-			    -DCMAKE_INSTALL_PREFIX=/opt/local/llvm \
-			    -DLLVM_TARGETS_TO_BUILD="ARM" \
-			    -DLLVM_TARGET_ARCH="ARM"
-	    fi
-	    ninja lld
-	    $MAKE_ME_ROOT ninja install-lld
-	    popd
+    if ! which rustup >/dev/null 2>&1; then
+      curl https://sh.rustup.rs -sSf | sh -s -- -y
+      source ~/.cargo/env
```
TriplEight: does it work equally being launched with sudo and from root?

riusricardo: mmm, I don't think so. That depends on the user you use to install Rust. The Rust installation command always installs in the home directory. So if a user runs as root, it will install rust in the root's home. However I think that is the expected behavior of the command.

PR #159, opened 2020-02-27
Add deprecation message to get-parity.io

s3krit:
Alerts users that parity-ethereum is now open-ethereum and provides a link to the new location


kirushik:
Maybe we can make that `sleep` adding another second for each day passing from now on?
So you know, in a year only the dumbest user won't notice that something's wrong...

PR #175, opened 2020-04-11
don't upgrade every brew package

lumir-mrkva:
this is fix of possible typo. upgrade of everything shouldn't be necessary - updating casks should be enough to get fresh brews for substrate dependencies.


lumir-mrkva:
[clabot:check]

andresilva:
I don't have a mac now to test this but it seems correct to me. Thanks for the PR.

PR #191, opened 2020-05-17
ink!: Add miri component and generally improve script

Robbepop:


dockerfiles/ink-ci-linux/Dockerfile
```diff
@@ -21,23 +21,28 @@ RUN	set -eux; \
...
+#  - miri: Rust interpreter with additional safety checks.
+#
+# We also need to install the wasm32-unknown-unknown target to test
+# ink! smart contracts compilation for this target architecture.
+#
+# Only Rust nightly builds supporting all of the above mentioned components
+# and targets can be used for this docker image.
+#
+# Visit https://rust-lang.github.io/rustup-components-history/ to see which
+# components are available for which nightly build.
```
TriplEight, line 36: maybe we should move some comments to the readme?

dockerfiles/ink-ci-linux/Dockerfile
```diff
@@ -21,23 +21,28 @@ RUN	set -eux; \
...
+	rustup default nightly-2020-05-14; \
+	# Install wasm32 target for this toolchain
 	rustup target add wasm32-unknown-unknown; \
-# install cargo tools
-	rustup component add rustfmt clippy; \
+	rustup component add rustfmt clippy miri; \
 	cargo install grcov rust-covfix; \
-	cargo install --git https://github.com/paritytech/cargo-contract; \
-# versions
+	cargo install cargo-contract; \
```
TriplEight, line 42: are you sure you want stable `cargo-contract`? Then it could be installed in the upper line, but I do not care

dockerfiles/ink-ci-linux/Dockerfile
Robbepop: i think the comments are better here since they are for the developers that are touching the file directly, not for the user of the dockerfile. it is neat to have information about how to update next to what you want to update in a file.

dockerfiles/ink-ci-linux/Dockerfile
Robbepop: what does that mean: upper line installment?
yes, stable `cargo-contract` would be nice. we are testing ink!, not `cargo-contract` here.

dockerfiles/ink-ci-linux/Dockerfile
```diff
@@ -21,23 +21,28 @@ RUN	set -eux; \
...
+#
+# Visit https://rust-lang.github.io/rustup-components-history/ to see which
+# components are available for which nightly build.
+	rustup default nightly-2020-05-14; \
+	# Install wasm32 target for this toolchain
 	rustup target add wasm32-unknown-unknown; \
-# install cargo tools
-	rustup component add rustfmt clippy; \
+	rustup component add rustfmt clippy miri; \
 	cargo install grcov rust-covfix; \
```
TriplEight, line 41:
```suggestion
	cargo install grcov rust-covfix cargo-contract; \
```

dockerfiles/ink-ci-linux/Dockerfile
TriplEight, line 42:
```suggestion
```

PR #192, opened 2020-05-18
Prep to deprecation in GitLab 13

TriplEight:
Before v13 we have to move from using `only/except` with templates, so trying to do it I stumbled upon a bug that their template [uses](https://gitlab.com/gitlab-org/gitlab/-/blob/68ba7914479b56bb04f5737f6e410772b3f70812/lib/gitlab/ci/templates/Security/Container-Scanning.gitlab-ci.yml#L28) `only`.
Had to workaround their [official](https://docs.gitlab.com/ee/user/application_security/container_scanning/#configuration) suggestion.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -22,22 +22,16 @@ variables:
 .build:                            &docker_build
   stage:                           build
   image:                           docker:stable
-  only:
-    refs:
-      - schedules
-      - web
-    kubernetes:                    active
+  rules:
+    - if:                          '$DOCKERIMAGE == $CI_JOB_NAME'
+    - changes:                     
+      - dockerfiles/${IMAGE_NAME}'
   tags:
     - kubernetes-parity-build
   services:
     - docker:dind
   environment:
     name:                          parity-build
-  before_script:
-    - test -z "$DOCKERIMAGE" && echo "DOCKERIMAGE must be defined" && exit 1
-  only:
-    variables:
-      - $DOCKERIMAGE == $CI_JOB_NAME
 
 # Push to Dockerhub
 .push_to_docker_hub:               &push_to_docker_hub
@@ -215,9 +209,10 @@ terraform:
 # in order to be scanned, EVERY docker build HAS to have $REGISTRY_PATH/$IMAGE_NAME:$IMAGE_TAG
 container_scanning:
   stage:                           test
-  only:
-    variables:
-      - $DOCKERIMAGE
+  rules:
+    # this is similar to only: variables: $x
+    - if:                          '$DOCKERIMAGE'
+      when:                        always
   variables:
     GIT_STRATEGY:                  fetch
     CI_APPLICATION_REPOSITORY:     $REGISTRY_PATH/$DOCKERIMAGE
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -15,7 +15,6 @@ variables:
   DOCKER_DRIVER:                   overlay2
   IMAGE_NAME:                      $CI_JOB_NAME
   IMAGE_TAG:                       latest
-  DOCKERFILE_DIR:                  $CI_JOB_NAME
   REGISTRY_PATH:                   paritytech
   CI_REGISTRY_PATH:                "registry.parity.io/parity/infrastructure/scripts"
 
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -22,22 +19,18 @@ variables:
 .build:                            &docker_build
   stage:                           build
   image:                           docker:stable
-  only:
-    refs:
```
TriplEight, line 26: @gabreal FYI, I'm removing this since CI is triggered only by $DOCKERFILES variable anyway

.gitlab-ci.yml
```diff
@@ -22,22 +19,18 @@ variables:
...
-      - schedules
-      - web
-    kubernetes:                    active
```
TriplEight, line 29: @gabreal FYI, I'm removing this since it is about if k8s is turned on for the project, it doesn't check the status of k8s. And "rules:" syntax does not have this status implemented.

.gitlab-ci.yml
```diff
@@ -22,22 +19,18 @@ variables:
...
+    - changes:
+      - 'dockerfiles/${IMAGE_NAME}/Dockerfile'
   tags:
     - kubernetes-parity-build
   services:
     - docker:dind
   environment:
     name:                          parity-build
-  before_script:
-    - test -z "$DOCKERIMAGE" && echo "DOCKERIMAGE must be defined" && exit 1
```
TriplEight, line 37: @gabreal FYI, I'm removing this since CI is triggered only by $DOCKERFILES variable anyway

.gitlab-ci.yml
```diff
@@ -214,17 +207,37 @@ terraform:
...
 container_scanning:
+  # Template does not work, had to reimplement its config from
+  # https://gitlab.com/gitlab-org/gitlab/blob/master/lib/gitlab/ci/templates/Security/Container-Scanning.gitlab-ci.yml
   stage:                           test
-  only:
-    variables:
-      - $DOCKERIMAGE
+  image:                           $SECURE_ANALYZERS_PREFIX/klar:$CS_MAJOR_VERSION
   variables:
+    SECURE_ANALYZERS_PREFIX:       "registry.gitlab.com/gitlab-org/security-products/analyzers"
```
General-Beck, line 215: MB move this VAR to CI variables?

.gitlab-ci.yml
```diff
@@ -214,17 +207,37 @@ terraform:
...
+    CS_MAJOR_VERSION:              2
+    CLAIR_DB_IMAGE_TAG:            "latest"
+    CLAIR_DB_IMAGE:                "$SECURE_ANALYZERS_PREFIX/clair-vulnerabilities-db:$CLAIR_DB_IMAGE_TAG"
     GIT_STRATEGY:                  fetch
     CI_APPLICATION_REPOSITORY:     $REGISTRY_PATH/$DOCKERIMAGE
     CI_APPLICATION_TAG:            $IMAGE_TAG # OR $IMAGE_DATE_TAG
     DOCKERFILE_PATH:               "dockerfiles/${DOCKERIMAGE}"
     DOCKER_USER:                   $Docker_Hub_User_Parity
```
General-Beck, line 223: already in CI Variables

.gitlab-ci.yml
```diff
@@ -50,7 +43,7 @@ variables:
         --tag $REGISTRY_PATH/$IMAGE_NAME:$IMAGE_DATE_TAG
         --tag $REGISTRY_PATH/$IMAGE_NAME:$IMAGE_TAG
         --file $IMAGE_NAME/Dockerfile .
-    - docker login -u "${Docker_Hub_User_Parity}" -p "${Docker_Hub_Pass_Parity}"
+    - docker login -u "${DOCKER_USER}" -p "${DOCKER_PASSWORD}"
```
gabreal, line 46: what about using a syntx like here: https://docs.docker.com/engine/reference/commandline/login/ along with https://docs.gitlab.com/ee/ci/variables/#custom-environment-variables-of-type-file?
the password varable will not be masked in the scripts (would be nice though)

.gitlab-ci.yml
```diff
@@ -22,22 +19,18 @@ variables:
...
+    - if:                          '$DOCKERIMAGE == $CI_JOB_NAME'
+    # triggers rebuild when Dockerfile is changed
+    - changes:
+      - 'dockerfiles/${IMAGE_NAME}/Dockerfile'
   tags:
     - kubernetes-parity-build
   services:
     - docker:dind
   environment:
     name:                          parity-build
```
gabreal, line 33: afaiu the environment block can also go, since the jobs don't need to have access to the kubernetes api.

TriplEight:
oof, I've just broke everything, reverting

PR #196, opened 2020-05-19
WIP: POC test substrate

TriplEight:


TriplEight:
Here's an example of how it's supposed to work:
https://gitlab.parity.io/parity/infrastructure/scripts/pipelines/93402
- image is build and pushed as `nightly` and `hash-date`
- image is used to buid substrate
- of OK, image is pushed as stable to be used by Substrate CI

PR #199, opened 2020-05-28
Contracts solang update

TriplEight:
using solang image to get the binary from


TriplEight:
damn, I didn't originate it from master

PR #200, opened 2020-05-28
contracts ci image change how solang is installed

TriplEight:
using `solang` image to get the binary from


Stefie:
See https://github.com/paritytech/pallet-contracts-waterfall/issues/77#issuecomment-635927621

PR #208, opened 2020-06-10
Remove rust-builder files

TriplEight:


wischli:
May I ask why the rust-builder files were removed? I am currently using the image for building a Substrate Node image.

TriplEight:
@wischli sure, I removed this one because prepared a [replacement](https://github.com/paritytech/scripts/tree/master/dockerfiles/ci-linux): `paritytech/ci-linux:production`. 

This new image is the same as the previous one, but built in two steps and tested against Substrate CI. It only reaches `:production` tag if Substrate tests are successfull.

Also, we are still pushing this same image to `parity/rust-builder:latest`, but will deprecate it soon.

PR #210, opened 2020-06-21
Update LLVM branch to latest

yuki-js:


yuki-js:
[clabot:check]

TriplEight:
Hi, can you provide us with any explanation why it should be updated? Does it even work with this version?

yuki-js:
LLVM branch name seems to be changed so cloning failed.
I could successfully compile it after changing it.

dockerfile: https://github.com/my-number/mynachain/blob/master/Dockerfile.arm
image: https://github.com/my-number/mynachain/packages/280247

TriplEight:
Seems [legit](https://github.com/rust-lang/rust/pull/67759/files). Thank you, @yuki-js for the contribution!

PR #219, opened 2020-07-29
docker push rust-builder image

fevo1971:
`rust-builder:latest` docker image was never pushed.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -283,6 +283,7 @@ ci-linux-production:
     - docker tag $REGISTRY_PATH/$IMAGE_NAME:staging $REGISTRY_PATH/$IMAGE_NAME:production
     - docker tag $REGISTRY_PATH/$IMAGE_NAME:staging $REGISTRY_PATH/$IMAGE_NAME:$IMAGE_DATE_TAG
     - docker tag $REGISTRY_PATH/$IMAGE_NAME:staging parity/rust-builder:latest
+    - docker push parity/rust-builder:latest
     - docker push $REGISTRY_PATH/$IMAGE_NAME:production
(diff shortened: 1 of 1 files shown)


gabreal:
not sure about the permissions though, needs to be verified on the first run.

TriplEight:
Is this tag needed somewhere? It's a long time gone and last time was mentioned as a temporary measure. I suggest removing the `docker tag` line.

gabreal:
I'm also fine with removing it entirely, not sure anymore where it was needed. do you @fevo1971 ?

.gitlab-ci.yml
```diff
@@ -283,6 +283,7 @@ ci-linux-production:
     - docker tag $REGISTRY_PATH/$IMAGE_NAME:staging $REGISTRY_PATH/$IMAGE_NAME:production
     - docker tag $REGISTRY_PATH/$IMAGE_NAME:staging $REGISTRY_PATH/$IMAGE_NAME:$IMAGE_DATE_TAG
     - docker tag $REGISTRY_PATH/$IMAGE_NAME:staging parity/rust-builder:latest
+    - docker push parity/rust-builder:latest
```
TriplEight, line 286:
```suggestion
```

TriplEight (changes requested):
I've searched the whole github a couple of months ago to make sure that no active projects are using `rust-builder` or `substrate-ci-linux` to transit to `ci-linux` safely and even did PRs in a couple of third-party projects (and obviously in ours as well) to rename the image.
I doubt that something's changed since that time. 
After this change, I can remove the older images from docker hub for good.

TriplEight:
this is a tiny bit outdated

PR #227, opened 2020-10-15
Print error message on build fail and point to guidelines

thiolliere:
This PR now prints some information when build fails.

I'm not sure about macos colors, but I think they are same.

I'm not sure about those changes, but sounds like an improvment to me

### DIFF
--- a/get-substrate.sh
+++ b/get-substrate.sh
@@ -78,11 +78,17 @@ if [[ "$1" == "--fast" ]]; then
 	echo "You can install manually by cloning the https://github.com/paritytech/substrate repo,"
 	echo "and using cargo to install 'substrate' and 'subkey' from the repo path."
 else 
+	BOLD_RED='\033[1;31m'
+	NO_COLOR='\033[0m'
+	ERROR_MSG="\n\n${BOLD_RED}Build failed:${NO_COLOR} This can be due to latest rust nightly \
+breaking substrate, for more precise guidelines go to \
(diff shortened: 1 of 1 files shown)


get-substrate.sh
```diff
@@ -78,11 +78,17 @@ if [[ "$1" == "--fast" ]]; then
 	echo "You can install manually by cloning the https://github.com/paritytech/substrate repo,"
 	echo "and using cargo to install 'substrate' and 'subkey' from the repo path."
 else 
+	BOLD_RED='\033[1;31m'
+	NO_COLOR='\033[0m'
+	ERROR_MSG="\n\n${BOLD_RED}Build failed:${NO_COLOR} This can be due to latest rust nightly \
+breaking substrate, for more precise guidelines go to \
```
TriplEight, line 84: Actually, we could run ` rustc --version` and ` rustc +nightly --version` [here](https://github.com/paritytech/substrate/blob/master/.gitlab-ci.yml#L417), save it to build artifacts and/or binary metadata to query in different places.
This script, for instance, could at least warn about the mismatch of the rust versions.

TriplEight:
@thiolliere is it still relevant?

TriplEight:
@thiolliere is it still relevant?

thiolliere:
The related command will still fail with annoyingly when substrate is not supported by latest nightly.
And ppl will get confused why the build failed.

But maybe this script is not recomanded anymore so it is fine.

PR #230, opened 2020-10-19
Add bash to awscli image

s3krit:
Due to [this](https://gitlab.com/gitlab-org/gitlab-runner/-/issues/1550) bug in gitlab-runner, a non-POSIX feature is used even when the shell is not `bash`. I {am using|will use} this image to push release-candidate builds to S3 on the cleanroom (see [this](https://gitlab.parity.io/s3krit/cleanroom/-/tree/gitlab-runner/ansible/roles/polkadot/files/gitlab-runner) branch or just ask me if you're curious about the progress of that), so run into this issue.

The actual issue is with the [`set -eo pipefail`](https://gitlab.com/gitlab-org/gitlab-runner/-/blob/master/shells/bash.go#L224) - some POSIX shells *do* seem to support this (or at least don't error when setting it), such as busybox's shell. So far I've only ran into `dash` that will actually error when attempting it.

### DIFF
--- a/dockerfiles/awscli/Dockerfile
+++ b/dockerfiles/awscli/Dockerfile
@@ -17,6 +17,7 @@ dockerfiles/awscli/README.md" \
 	io.parity.image.created="${BUILD_DATE}"
 
 RUN pip install awscli
+RUN apk add bash
 
 COPY utility/awscli-config /etc/aws.config
 ENV AWS_CONFIG_FILE /etc/aws.config
--- a/dockerfiles/awscli/Dockerfile
+++ b/dockerfiles/awscli/Dockerfile
@@ -17,7 +17,7 @@ dockerfiles/awscli/README.md" \
 	io.parity.image.created="${BUILD_DATE}"
 
 RUN pip install awscli
-RUN apk add bash
+RUN apk add --no-cache bash
 
 COPY utility/awscli-config /etc/aws.config
 ENV AWS_CONFIG_FILE /etc/aws.config


General-Beck (approved):

TriplEight (approved):
:+1:

gabreal:
I don't mind adding it if it's needed, would suggest to use `apk add --no-cache bash` though.

actually i build the image specifically for using it in the gitlab ci and never encountered any issues, also cannot find the usage of the image in your branch.

s3krit:
>I don't mind adding it if it's needed, would suggest to use apk add --no-cache bash though.

Sure thing, will do

> actually i build the image specifically for using it in the gitlab ci and never encountered any issues, also cannot find the usage of the image in your branch.

Yeah I don't know why it doesn't manifest when you run it as a 'normal' (i.e., not a custom) executor. Though the person that reported the bug originally did so before custom executors were a thing.  I never call the non-POSIX line directly, it is called by gitlab-runner itself in [this](https://gitlab.com/gitlab-org/gitlab-runner/-/blob/master/shells/bash.go#L224) line of shell-detection code.

gabreal:
so you are using the `awscli` image with podman?

gabreal (approved):
consider `--no-cache` option

s3krit:
> so you are using the `awscli` image with podman?

The code's not quite ready yet but yes - the intent is to run the build-linux-release and publish-s3-release jobs from the cleanroom server for release-candidate and release binaries.

PR #231, opened 2020-10-22
temporary override for contracts' CI nightly toolchain

TriplEight:
for https://github.com/paritytech/cargo-contract/pull/86

### DIFF
--- a/dockerfiles/contracts-ci-linux/Dockerfile
+++ b/dockerfiles/contracts-ci-linux/Dockerfile
@@ -39,12 +39,12 @@ RUN set -eux; \
 # use minimum components
 	rustup set profile minimal; \
 # install nightly Rust
-	rustup install nightly; \
+	rustup install nightly-2020-10-01; \
 # install wasm toolchain
-	rustup target add wasm32-unknown-unknown --toolchain nightly; \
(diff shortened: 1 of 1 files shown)


ascjones (approved):
Thanks!

dockerfiles/contracts-ci-linux/Dockerfile
```diff
@@ -39,12 +39,12 @@ RUN set -eux; \
 # use minimum components
 	rustup set profile minimal; \
 # install nightly Rust
-	rustup install nightly; \
+	rustup install nightly-2020-10-01; \
```
ascjones, line 42: Hm I think we might need to add `rustup default nightly-2020-10-01`.

Failure: https://gitlab.parity.io/parity/cargo-contract/-/jobs/691564#L55

TriplEight: It would be overridden by [RUSTUP_TOOLCHAIN](https://github.com/paritytech/cargo-contract/blob/b3913161342eae2f631de5d437eecfe55fea3ac4/.gitlab-ci.yml#L20), should remove that line as well.

PR #235, opened 2020-11-02
Enable binaryen-sys build in ink! container

cmichi:
Escalated from https://github.com/paritytech/cargo-contract/pull/95.

### DIFF
--- a/dockerfiles/ink-ci-linux/Dockerfile
+++ b/dockerfiles/ink-ci-linux/Dockerfile
@@ -18,6 +18,9 @@ dockerfiles/ink-ci-linux/README.md" \
 	io.parity.image.created="${BUILD_DATE}"
 
 RUN	set -eux; \
+# Necessary for building binaryen-sys, which is part of cargo-contract's
+# binaryen dependency
+  export CXX="/usr/bin/clang++-8" \
 # The supported Rust nightly version must support the following components
 # to allow for a functioning CI pipeline:
 #
--- a/dockerfiles/ink-ci-linux/Dockerfile
+++ b/dockerfiles/ink-ci-linux/Dockerfile
@@ -17,10 +17,11 @@ dockerfiles/ink-ci-linux/README.md" \
 	io.parity.image.revision="${VCS_REF}" \
 	io.parity.image.created="${BUILD_DATE}"
 
-RUN	set -eux; \
 # Necessary for building binaryen-sys, which is part of cargo-contract's
 # binaryen dependency
-  export CXX="/usr/bin/clang++-8" \
+ENV CXX="/usr/bin/clang++-8"
+
+RUN	set -eux; \
 # The supported Rust nightly version must support the following components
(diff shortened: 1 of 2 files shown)


dockerfiles/ink-ci-linux/Dockerfile
```diff
@@ -18,6 +18,9 @@ dockerfiles/ink-ci-linux/README.md" \
 	io.parity.image.created="${BUILD_DATE}"
 
 RUN	set -eux; \
```
TriplEight, line 20:
```suggestion
# Necessary for building binaryen-sys, which is part of cargo-contract's
# binaryen dependency
ENV CXX="/usr/bin/clang++-8"

RUN	set -eux; \
```

dockerfiles/ink-ci-linux/Dockerfile
```diff
@@ -18,6 +18,9 @@ dockerfiles/ink-ci-linux/README.md" \
...
+# Necessary for building binaryen-sys, which is part of cargo-contract's
```
TriplEight, line 21:
```suggestion
# Necessary for building binaryen-sys, which is part of cargo-contract's
```

dockerfiles/ink-ci-linux/Dockerfile
```diff
@@ -18,6 +18,9 @@ dockerfiles/ink-ci-linux/README.md" \
...
+# binaryen dependency
```
TriplEight, line 22:
```suggestion
# binaryen dependency
```

dockerfiles/ink-ci-linux/Dockerfile
```diff
@@ -18,6 +18,9 @@ dockerfiles/ink-ci-linux/README.md" \
...
+  export CXX="/usr/bin/clang++-8" \
```
TriplEight, line 23:
```suggestion
  export CXX="/usr/bin/clang++-8" \
```

dockerfiles/ink-ci-linux/Dockerfile
TriplEight: environment variable in Dockerfiles are set via `ENV`

TriplEight (changes requested):

TriplEight:
That doesn't work with some other dependencies, could there be any other way instead of changing the linker for everything?

cmichi:
@TriplEight Ok, hmm, the only way I see is putting it back in the `.gitlab-ci.yml` from `cargo-contracts`, so that it affects only what it strictly needs to affect :/.

TriplEight:
here's everything we have in the CI image:
```
root@cb1df0df5555:/builds# ls /usr/bin/ | grep clang
clang
clang++
clang++-7
clang-7
clang-cpp-7
```

And you've mentioned there should be C++14

TriplEight:
Done in https://github.com/paritytech/scripts/pull/237

PR #237, opened 2020-11-04
Update solang and canvas-node in contracts CI image

TriplEight:
> Can't be merged until https://github.com/paritytech/canvas-node/issues/21 is resolved, this will break the `test` job in `cargo-contracts` CI. Other than that, `fmt` and `build` are successful.

Temporary workaround: https://github.com/paritytech/cargo-contract/pull/114

- `solang` update to `v0.1.2`, newer versions did not work
- `canvas`-node binary included as an artifact from the latest master
- `clang-8` set as a linker
- `python3` set as default

Closes paritytech/ci_cd#1 
Closes paritytech/ci_cd#18

### DIFF
--- a/dockerfiles/contracts-ci-linux/Dockerfile
+++ b/dockerfiles/contracts-ci-linux/Dockerfile
@@ -20,6 +20,8 @@ dockerfiles/contracts-ci-linux/README.md" \
 
 WORKDIR /builds
 
+ENV CXX="/usr/bin/clang++-8"
+
 # copy llvm and yarn repo key
 COPY utility/yarn.key /etc/apt/trusted.gpg.d/debian-yarn.gpg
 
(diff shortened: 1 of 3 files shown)


TriplEight:
@ascjones @cmichi @Robbepop  how to proceed with it? Is there any progress with https://github.com/paritytech/canvas-node/issues/21

ascjones:
I thought that paritytech/canvas-node#21 is solved by this PR, because the binary no longer is compiled by this image, instead it is now downloaded from the result of the canvas-node CI which currently runs successfully.

TriplEight:
@ascjones sort-ta. There's no problem to `build` `cargo-contract` within this image, I specifically tested this.
But the `test` job of `cargo-contract` CI would fail on the same `error: could not compile sp-arithmetic`.

ascjones:
Summary:

- `cargo-contract` depends on the latest release of substrate `2.0.0` on crates.io
- `substrate 2.0.0` fails to build with `nightly` > `2020-10-01`
- This PR produces an image which contains a newer (incompatible) version of nightly

Therefore this PR should not be merged until a new version of substrate is available on crates.io which can compile with the latest nightly, and `cargo-contract` has been upgraded to depend on that version.

cmichi (approved):

PR #238, opened 2020-11-05
Trigger contracts CI to test the image against master

TriplEight:
Contracts CI image is now tested against contract code every time the image is updated (nightly).

Pipeline 
1. Scripts CI builds `contracts-ci-linux:staging`
2. Runs container scan and
    Triggers the downstream Contracts CI pipeline for the master branch [to be executed in](https://gitlab.parity.io/parity/cargo-contract/-/jobs/732567#L4) `contracts-ci-linux:staging` image
3. If the test is green, `contracts-ci-linux:staging` gets published as `contracts-ci-linux:production`
    In the other case, It's easier for the developers to see the failed pipeline in their own CI rather than look into this one (which no one does).

A good example of the pipeline: https://gitlab.parity.io/parity/infrastructure/scripts/-/pipelines/115699

image

Companion PR: https://github.com/paritytech/cargo-contract/pull/115

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -280,6 +280,9 @@ ci-linux-test:
 
 contracts-ci-linux-test:
   stage:                           test
+  only:
+    variables:
+      - $IMAGE_NAME == "contracts-ci-linux"
   needs:
     - job:                         contracts-ci-linux
       artifacts:                   false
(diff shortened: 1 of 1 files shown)


TriplEight:
@ascjones @cmichi check this out

General-Beck (approved):

PR #242, opened 2020-11-24
WIP: new image for sccache CI

TriplEight:
Closes paritytech/ci_cd#73

### DIFF
--- a/dockerfiles/sccache-ci-linux/Dockerfile
+++ b/dockerfiles/sccache-ci-linux/Dockerfile
@@ -1,0 +1,101 @@
+FROM ubuntu:groovy
+
+ARG VCS_REF=master
+ARG BUILD_DATE=""
+ARG REGISTRY_PATH=paritytech
+
+# metadata
+LABEL summary="CI image with all dependencies for sccache compilation." \
+	name="${REGISTRY_PATH}/sccache-ci-linux" \
+	maintainer="devops-team@parity.io" \
+	version="1.0" \
+	description="libssl-dev, clang, libclang-dev, lld, cmake, make, git, pkg-config \
+curl, time, rhash, rust stable, rust nightly, sccache" \
+	io.parity.image.vendor="Parity Technologies" \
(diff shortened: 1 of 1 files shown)


drahnr (approved):

dockerfiles/sccache-ci-linux/Dockerfile
```diff
@@ -0,0 +1,85 @@
+FROM ubuntu:groovy
```
drahnr, line 1: :+1:

dockerfiles/sccache-ci-linux/Dockerfile
```diff
@@ -0,0 +1,85 @@
...
+	CARGO_HOME=/usr/local/cargo \
+	PATH=/usr/local/cargo/bin:/usr/local/cuda-11.1/bin:$PATH \
+	CC=clang \
+	CXX=clang \
+    LD_LIBRARY_PATH=/usr/local/cuda-11.1/lib64 \
+    DEBIAN_FRONTEND=noninteractive \
+    TERM=xterm-256color
+
+# install tools and dependencies
+RUN set -eux; \
```
drahnr, line 37: With newer versions of podman / docker should be able to fuse the layers, so the spaghetti/chain wars should not be necessary anymore but changes can take advantage of cached layers :)

TriplEight: Yeah, the thing is we ignore this layer cache on the CI images builds.
And the reason is I didn't update myself about some bugs that were in docker layer caching (like if the line in Dockerfile that `COPY`es the file from local doesn't change, but the file occasionally got changed, still the old file version can be taken from the cache).

TriplEight:
However, it's still WIP, as I didn't yet find some time to figure out what has to be configured to satisfy the dist tests.

TriplEight:
I think it's already a good substitution for the current CI image. Further work on it will require the image to be in the CI loop.

PR #258, opened 2021-02-15
CI: trigger substrate CI on image test

TriplEight:
For the better maintainability and observability it's good to trigger the repo itself, so developers will know what stops the CI image from being updated.

Overrides `RUSTFLAGS` to avoid failures on `-Dwarnings`, in this case it's not needed.

Merge after `CI_IMAGE` is introduced to substrate CI repo.


gabreal (approved):
lgtm

PR #277, opened 2021-03-26
dockerfiles: add sops docker image

PierreBesson:
Add a docker image containing sops for encrypting and decrypting secrets contained in files using an encryption-as-a-service provider (KMS) such as Vault.
The vault binary is added for convenience to preform login operations on Vault and get the vault token that will be passed to sops.


TriplEight (approved):
:+1: 
However, there are already enough similar builds in CI config so they could use an anchor.

PR #288, opened 2021-04-07
Install `geckodriver` and `firefox` for UI tests

cmichi:


dockerfiles/ink-waterfall-ci/Dockerfile
```diff
@@ -25,6 +25,19 @@ ENV DEBIAN_FRONTEND=noninteractive
 
 RUN	set -eux; \
 	apt-get -y update && \
+	apt-get install -y --no-install-recommends \
+# `firefox` is needed to simulate interactions with the `canvas-ui`
+	firefox && \
```
TriplEight, line 30: Not against firefox at all, just to make a point.
We use chromium at substrate CI:
https://github.com/paritytech/scripts/blob/master/dockerfiles/ci-linux/Dockerfile#L26

cmichi: Okay good to know, but the library which I'm using supports only geckodriver, which is tied to firefox.

TriplEight: cool

TriplEight (approved):

PR #295, opened 2021-05-17
Integrate `rand-extension` into `ink-waterfall`

cmichi:
Follow-up for https://github.com/paritytech/ink/pull/793.

### DIFF
--- a/dockerfiles/ink-waterfall-ci/Dockerfile
+++ b/dockerfiles/ink-waterfall-ci/Dockerfile
@@ -52,13 +52,27 @@ RUN	set -eux; \
 # `--locked` ensures the project's `Cargo.lock` is used.
 	cargo install --git https://github.com/paritytech/canvas-node.git \
 		--locked --branch master && \
+# We additionally install the `canvas-node` as `canvas-rand-extension`.
+# This installation though is a modified `canvas-node`, so that ink!'s
+# `rand-extension` chain extension example is included in the runtime.
+# This enables us to test in the waterfall that the `rand-extension`
+# integration with Substrate still works.
+	git clone --depth 1 https://github.com/paritytech/canvas-node.git && \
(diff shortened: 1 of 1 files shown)


dockerfiles/ink-waterfall-ci/Dockerfile
```diff
@@ -52,13 +52,27 @@ RUN	set -eux; \
...
+# We additionally install the `canvas-node` as `canvas-rand-extension`.
+# This installation though is a modified `canvas-node`, so that ink!'s
+# `rand-extension` chain extension example is included in the runtime.
+# This enables us to test in the waterfall that the `rand-extension`
+# integration with Substrate still works.
+	git clone --depth 1 https://github.com/paritytech/canvas-node.git && \
+	curl -s https://raw.githubusercontent.com/paritytech/ink/master/examples/rand-extension/runtime/chain-extension-example.rs \
+		>> canvas-node/runtime/src/lib.rs && \
+	sed -i 's/type ChainExtension = ();/type ChainExtension = FetchRandomExtension;/g' canvas-node/runtime/src/lib.rs && \
+	sed -i 's/name = "canvas"/name = "canvas-rand-extension"/g' canvas-node/node/Cargo.toml && \
```
TriplEight, line 64: I see the potential non-documented issues with this custom installation, maybe it makes sense to create a feature flag instead? That way it would be better supported.

cmichi: You mean a feature flag in the `canvas-node` code base, right? I had thought about this as well, but the thing is that everything regarding `ink-waterfall` is just meant for testing and I would rather not have anything regarding ink! examples in our production code base for `canvas-node`.

TriplEight: And this kind of testing is only meant to be internal, right?

cmichi: Yup!

TriplEight (approved):

PR #302, opened 2021-05-26
Add non-root user for the ops tooling images

rcny:
Relates to paritytech/ci_cd#92.

This PR adds non-root user only for ops tooling images as per my comment [here](https://github.com/paritytech/ci_cd/issues/92#issuecomment-847926377).

Please review the changes. After the merge, I'll rebuild the affected images.


gabreal (approved):
:+1: for making the containers rootless, don't see the need to install bash everywhere though, but i leave that up to you.

TriplEight (approved):
`shadow` is for `groupadd/useradd`, `bash` is just handy, LGTM

PR #304, opened 2021-05-27
Non-root user for most of the CI images

rcny:
Relates to paritytech/ci_cd#92.

### DIFF
--- a/dockerfiles/base-ci-linux/Dockerfile
+++ b/dockerfiles/base-ci-linux/Dockerfile
@@ -59,7 +59,10 @@ RUN set -eux; \
 # apt clean up
 	apt-get autoremove -y; \
 	apt-get clean; \
-	rm -rf /var/lib/apt/lists/*
+  rm -rf /var/lib/apt/lists/*; \
+# add non-root user
+  groupadd -g 1000 nonroot; \
+  useradd -u 1000 -g 1000 -s /bin/bash -m nonroot
 # cache handler
 ENV	RUSTC_WRAPPER=sccache \
 # show backtraces
--- a/dockerfiles/base-ci/Dockerfile
+++ b/dockerfiles/base-ci/Dockerfile
@@ -62,7 +62,10 @@ RUN set -eux; \
 # apt clean up
 	apt-get autoremove -y && \
 	apt-get clean && \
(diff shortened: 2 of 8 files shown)


dockerfiles/base-ci-linux/Dockerfile
```diff
@@ -59,7 +59,10 @@ RUN set -eux; \
 # apt clean up
 	apt-get autoremove -y; \
 	apt-get clean; \
-	rm -rf /var/lib/apt/lists/*
+  rm -rf /var/lib/apt/lists/*; \
+# add non-root user
+  groupadd -g 1000 nonroot; \
+  useradd -u 1000 -g 1000 -s /bin/bash -m nonroot
```
TriplEight, line 65: both `base-ci-linux` and `base-ci` are designed to be used as the base layers for the other images, other than that they are not used.
So AFAIU you do the preparation part here (for DRY sakes) and actual user's called from this boilerplate in the end images, am I right?

dockerfiles/base-ci/Dockerfile
```diff
@@ -62,7 +62,10 @@ RUN set -eux; \
 # apt clean up
 	apt-get autoremove -y && \
 	apt-get clean && \
-	rm -rf /var/lib/apt/lists/*
+  rm -rf /var/lib/apt/lists/* && \
+# add non-root user
+  groupadd -g 1000 nonroot && \
+  useradd -u 1000 -g 1000 -s /bin/bash -m nonroot
```
TriplEight, line 68: I've read somewhere that it's generally safer to use ID > 10k to not to accidentally collapse with something, does it worth it?

Another source of "best practices" was even claiming that IDs are obsolete, are you playing safe here or is it better this way?

dockerfiles/base-ci-linux/Dockerfile
rcny: Yes, that's right.

dockerfiles/base-ci/Dockerfile
TriplEight: This is valid because we output caches and artifacts that are reused by the other containers and CI processes.

TriplEight (approved):

TriplEight:
I think this deserves a little doc.

PR #321, opened 2021-07-28
Add `wipe-inactive-runners` script

rcny:
```
Usage: wipe-inactive-runners [options]
        --tag RUNNER TAG             Which runner's tag to clean up
        --dry-run                    Just output what will be removed
```


sergejparity (approved):
Looks good to me

radupopa2010 (approved):
I can't ruby, I am more a pythin guy :).  It cleaned up the not connected runnes, so good job

TriplEight (approved):
LGTM. Although, we should think about moving these scripts somewhere into gitlab.

PR #323, opened 2021-08-05
Install `rustfmt` for stable

cmichi:
We are using `stable` for `canvas-node` and want to include `rustfmt` there.

CI fails currently with this:
```
$ cargo fmt --verbose --all -- --check
error: 'cargo-fmt' is not installed for the toolchain 'stable-x86_64-unknown-linux-gnu'
To install, run `rustup component add rustfmt`
```
https://gitlab.parity.io/parity/canvas-node/-/jobs/1049080


cmichi:
I'm closing this, since we decided to got for `cargo +nightly fmt` instead.

PR #331, opened 2021-08-24
Compile `cargo-spellcheck` with `stable` for `ink-ci-linux`

rcny:


dockerfiles/ink-ci-linux/Dockerfile
```diff
@@ -71,8 +71,15 @@ RUN	set -eux; \
...
 # We require `grcov` for coverage reporting and `rust-covfix` to improve it.
 # We require `cargo-spellcheck` for automated spell-checking
-	cargo install grcov rust-covfix xargo cargo-spellcheck && \
+	cargo install grcov rust-covfix xargo && \
 	cargo install cargo-contract && \
+
+# ICE with `cargo-spellcheck` on some 2021-08-xx nightlies:
+# https://gitlab.parity.io/parity/infrastructure/scripts/-/jobs/1078743#L1044
+# https://github.com/rust-lang/rust/issues/88309
+# Compile with stable for now
```
bkchr, line 80:
```suggestion
# Stuff that compiles on stable.
```

bkchr (approved):

KiChjang (approved):

dockerfiles/ink-ci-linux/Dockerfile
rcny: So we can compile it with `stable` and leave it be?

bkchr: I don't see any reason why we should not be able to do this ;)

PR #338, opened 2021-09-13
dockerfiles: update pinned nightly to 2021-09-13

TriplEight:
It's used in https://github.com/paritytech/polkadot/pull/3807 and https://github.com/paritytech/substrate/pull/9637 so if they are green, this may be merged.


rcny:
Should we update nightly now?

rcny (approved):

PR #352, opened 2021-10-22
Introduce polkadot-musl-builder

joao-paulo-parity:
Note that I will not able to follow up on the reviews for this PR because I'm going on vacation. I put this together before leaving in hopes it'll be easier for someone else to take over this work.

Additionally https://github.com/joao-paulo-parity/polkadot/blob/musl/Dockerfile is the link of where the script was ported from in case something got lost in translation.

Note: As documented in the script, I'm defaulting `POLKADOT_BRANCH` to `v0.9.9-1` because [lots of jemalloc undefined symbol errors](https://github.com/paritytech/scripts/files/7399902/tikv-jemalloc-link-error.txt) appeared after the switch tikv-jemallocator fork (https://github.com/paritytech/polkadot/pull/3897). Good news is that it's a single error which only happens during the very last step, when it's assembling the binary, and apparently does not seem to be difficult to solve - it might require an upstream PR on https://github.com/tikv/jemallocator though.

related to https://github.com/paritytech/release-engineering/issues/193

### DIFF
--- a/polkadot-musl-builder/README.md
+++ b/polkadot-musl-builder/README.md
@@ -1,0 +1,63 @@
+# Introduction
+
+[polkadot-musl-builder](./polkadot-musl-builder.sh) is script which builds a
+statically-linked Polkadot using musl.
+
+Note: This script is not self-contained! It's assumed that it will be ran in a
+clean slate container where we'll be free to mess around with the system's
+files and potentially leave the system's state broken in case of failures. You
+should not use it directly in your system.
+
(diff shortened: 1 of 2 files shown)


TriplEight (approved):

rcny:
Even that it's stale and kinda outdated, IMO this contribution is important, I'll keep it open to be picked up by someone maybe sometime in the future.

_cc_ @paritytech/release-engineering

PR #360, opened 2021-11-29
Add Ruby and `MestreLion/git-tools`

rcny:


alvicsam:
Why do we need ruby in the image?

rcny:
> Why do we need ruby in the image?

For future `cargo-cache-stash-update` and its consumers related routines.

alvicsam (approved):

PR #368, opened 2021-12-15
cachepot-ci: Add bubblewrap for dist compiles/tests

Xanewok:
Refs paritytech/cachepot#128 paritytech/cachepot#114

Otherwise we can't run sandboxed builds in our dist test suite, see https://gitlab.parity.io/parity/cachepot/-/jobs/1291693#L2738 for example

cc @drahnr


drahnr:
So the idea would be to call `unshare` wint `CLONE_NEWNS` map the user for file access, and then call bubble wrap as before, correct?

Xanewok:
At least for now, yeah.

drahnr (approved):

PR #370, opened 2021-12-20
Add benchmarks image

alvicsam:
This PR adds `paritytech/benchmarks` image.

In the issue https://github.com/paritytech/ci_cd/issues/275 was a request to check benchmarks and create a github issue if they exceeds threshold 20%

I created a script that does that but also decided to put this into a docker image so we can reuse it in other projects

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -140,6 +140,11 @@ tools:
   script:
     - *push_to_docker_hub
 
+benchmarks:
+  <<:                              *docker_build
+  script:
+    - *push_to_docker_hub
+
 query-exporter:
   <<:                              *docker_build
   script:
--- a/dockerfiles/benchmarks/.gitignore
+++ b/dockerfiles/benchmarks/.gitignore
@@ -1,0 +1,2 @@
+output.txt
+output_redacted.txt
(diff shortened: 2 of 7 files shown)


radupopa2010 (commented):
I have some improvement suggestions. The only hard thing would be point 3 in the list below. 
Size of output file should be small.

If you tested and everything works, is OK for me.

1. Why not using python for `dockerfiles/benchmarks/push_bench_results.sh` also?
If you have enough time, I would suggest using python for both scripts.

2. For parsing arguments and the config section of the python script I would suggest:
https://docs.python.org/3/library/argparse.html
https://docs.python.org/3/howto/argparse.html

3. How big is this file `bench_results = output_file.readlines()` and how much can grow?
Here you are reading the lines in memory;
A better approach would be:
```
with open(results_file, 'r') as file_handle:
    for line in file_handle:
        print(line)  # Do stuff with line
````

4. Body under  `if __name__ == "__main__":` is a bit too large and hard to read the intention.
Add one or two more functions to make it easier to understand what you are trying to do
Checkout chapter 3 in   https://github.com/ontiyonke/book-1/blob/master/%5BPROGRAMMING%5D%5BClean%20Code%20by%20Robert%20C%20Martin%5D.pdf

alvicsam:
@radupopa2010 thanks!

1. This was already written and I just added it as is. I'll rewrite it in the future.
2. I have only one argument, I wasn't sure that it worth to add `argparse`
3. The file is relatively small (currently about 120 lines, 8Kb) and I'm not sure that it'll grow much bigger
4. Okay, thx for recommendation, I'll think how can I improve that

radupopa2010 (approved):

rcny (approved):

PR #413, opened 2022-05-05
rpm: add .rpmmacros to image

coderobe:
see https://github.com/paritytech/cleanroom/pull/26#issuecomment-1118519963


chevdor (changes requested):
Using `COPY` would be simpler

chevdor (approved):

EgorPopelyaev (approved):

PR #416, opened 2022-05-19
Install diener from paritytech repository

joao-paulo-parity:
Alternative to https://github.com/paritytech/pipeline-scripts/pull/46/files#diff-df602127cd49392d5fb1da82c27351842395a47e788e0c6a77cf52d690ed1d2aR500 where instead of installing `diener` for every `check-dependent-*` job, we instead bake the binary into the image. That would be better than having to download and install it for each job.

Related to https://github.com/bkchr/diener/issues/15

### DIFF
--- a/dockerfiles/ci-linux/Dockerfile
+++ b/dockerfiles/ci-linux/Dockerfile
@@ -29,7 +29,8 @@ RUN set -eux && \
 	rustup target add wasm32-unknown-unknown --toolchain nightly && \
 	# install cargo tools
 	cargo install cargo-web wasm-pack cargo-deny cargo-spellcheck && \
-	cargo install --version 0.4.2 diener && \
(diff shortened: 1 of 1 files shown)


chevdor:
Another option that could benefit other CI scripts would be to propose a PR so diener releases include a deb.
There are several benefits, one of them being we can drop installing the rust toolchain, which is in some cases possible, and save lots of disk and build time.

joao-paulo-parity:
> propose a PR so diener releases include a deb

IMO better than a deb would be a statically-linked PIE binary. In fact we already have a project like that https://gitlab.parity.io/parity/opstooling/gitlab-guard/-/blob/957b3fd4a51dbc01daf5830b51d7769f03ceb4d2/build#L6-L9 so I think the build command could simply be copied for starters.

bkchr:
Proper solution is here: https://github.com/paritytech/cumulus/pull/1344

PR #427, opened 2022-06-02
DO NOT MERGE (yet) - Fix gnupg/rpm/deb containers

chevdor:
- rename `nonroot` as `user`
- fix acl: some of the files were not owned by the nonroot user
- add gpg2 symlink (required)

From the readme:
> Used as base for tooling that requires gnupg. GPG usually requires working with the gpg-agent.
> Using the gpg-agent in a rootless context can be challenging as you will need to align the UID
> in the container with the UIDs of your local system.
> 
> While we could make an image with UID that could be passed as ARG, this will likely always endup
> being the wrong UID. For this reason, this image is creating the users at runtime.
> By default, the UID is `1000` but you may customize it using the `LOCAL_USER_ID` environment variable.
> This will allow downstream images such as `paritytech/rpm` and `paritytech/deb` to be ran with the "right"
> UID and allows mapping the gpg-agent socket for the right UID.

This PR fixes the rootless rpm container calls as long as called like:

```
podman run  --rm -it \
   -v "/run/user/$(id -u)/gnupg/S.gpg-agent:/home/user/.gnupg/S.gpg-agent" \
   -v "$PWD/polkadot-0.9.23-1.x86_64.rpm:/home/user/polkadot-0.9.23-1.x86_64.rpm" \
   -v "$HOME/.gnupg/pubring.kbx:/home/user/.gnupg/pubring.kbx" \
    chevdor/rpm  \
   "rpm --addsign polkadot-0.9.23-1.x86_64.rpm"
```

### DIFF
--- a/dockerfiles/deb/Dockerfile
+++ b/dockerfiles/deb/Dockerfile
@@ -20,4 +20,5 @@ LABEL summary="DEB packaging/signing toolchain" \
 USER root
 RUN apt-get install -yq --no-install-recommends reprepro
 
-USER nonroot:nonroot
+ENTRYPOINT [ "/usr/local/bin/entrypoint.sh" ]
+CMD ["reprepro --version", "/bin/bash"]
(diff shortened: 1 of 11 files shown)


chevdor:
The RPM part can be dropped once https://github.com/paritytech/polkadot/pull/6595 is merged

chevdor:
We no longer support RPM so those changes and the rpm container could be dropped but the other chanegs may still be relevant.

PR #444, opened 2022-07-20
contracts-ci-linux: fix substract-contracts-node binary download URL

xermicus:
The download URL is incorrect. In current `:latest` and `:production` images the `substrate-contracts-node` file is just the HTML error page instead of the actual binary :grin: 

I will use this in solang integration tests

### DIFF
--- a/dockerfiles/contracts-ci-linux/Dockerfile
+++ b/dockerfiles/contracts-ci-linux/Dockerfile
@@ -49,7 +49,7 @@ RUN set -eux; \
 	cargo install cargo-dylint dylint-link && \
 	cargo install cargo-contract && \
 # download the latest `substrate-contracts-node` binary
-    curl -L -o substrate-contracts-node.zip 'https://gitlab.parity.io/parity/mirrors/substrate-contracts-node/-/jobs/1685841/artifacts/download?job=bud-linux' && \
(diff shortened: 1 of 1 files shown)


dockerfiles/contracts-ci-linux/Dockerfile
```diff
@@ -49,8 +49,10 @@ RUN set -eux; \
 	cargo install cargo-dylint dylint-link && \
 	cargo install cargo-contract && \
 # download the latest `substrate-contracts-node` binary
-	curl -L "https://gitlab.parity.io/parity/substrate-contracts-node/-/jobs/artifacts/master/raw/artifacts/substrate-contracts-node-linux/substrate-contracts-node?job=build-linux" \
-		-o /usr/local/cargo/bin/substrate-contracts-node && \
+    curl -L -o substrate-contracts-node.zip 'https://gitlab.parity.io/parity/mirrors/substrate-contracts-node/-/jobs/1685841/artifacts/download?job=bud-linux' && \
```
alvicsam, line 52: It's not a good choice to use artifacts from a job because we clean artifacts and job logs 1-2 times in a year.  If it's urgent we can merge it like this, if not I can fix it

alvicsam, line 52:
```suggestion
    curl -L -o substrate-contracts-node.zip 'https://gitlab.parity.io/parity/mirrors/substrate-contracts-node/-/jobs/artifacts/main/download?job=build-linux' && \
```
still need to fix the link :)

radupopa2010 (approved):

PR #451, opened 2022-08-23
Add cargo-hack to ci-linux

alvicsam:
As discussed in https://github.com/paritytech/jsonrpsee/pull/860 cargo-hack should be added to `ci-linux`

cc @lexnv


rcny (approved):

lexnv:
Thanks a lot for adding `cargo-hack` here!

PR #453, opened 2022-08-24
Use stable toolchain as default for ink! related CI images

HCastano:
After https://github.com/paritytech/cargo-contract/pull/698 we're able to build contracts
using the stable toolchain instead of nightly.

This updates our CI images to use stable as the default toolchain.

### DIFF
--- a/dockerfiles/contracts-ci-linux/Dockerfile
+++ b/dockerfiles/contracts-ci-linux/Dockerfile
@@ -23,7 +23,6 @@ WORKDIR /builds
 ENV SHELL /bin/bash
 ENV DEBIAN_FRONTEND=noninteractive
 
-# install tools and dependencies
 RUN set -eux; \
 	apt-get -y update && \
 	apt-get install -y --no-install-recommends \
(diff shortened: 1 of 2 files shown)


athei:
Won't this break the ink! CI? Building ink! itself (when not done by `cargo contract`) still needs nightly or `RUSTC_BOOSTRAP=1`.

HCastano:
You're right, I got too excited 😆 This change _could_ still work but we'd need to manually set `+nightly` to the places where we build ink! for Wasm. Tomorrow problem though, haha (so don't merge pls)

athei:
> This change could still work but we'd need to manually set +nightly to the places where we build ink! for Wasm.

I would say merge this but change the ink! CI to put `RUSTC_BOOTSTRAP=1` into the environment. Cause we should build on stable in our CI when we expect the same thing from our users.

HCastano:
> I would say merge this but change the ink! CI to put RUSTC_BOOTSTRAP=1 into the environment. Cause we should build on stable in our CI when we expect the same thing from our users.

Okay that makes sense. We can also use  `RUSTC_BOOTSTRAP=1` when doing our Clippy-Wasm checks which should save us some headaches [^1]. We will still need to use `+nightly` when running `cargo fmt` since we're using some config settings which they consider unstable.

[^1]: I've thought about this a bit and I'm not sure we want to do this, since contributors to ink! will have to use `RUSTC_BOOTSTRAP=1` locally to check that Clippy passes. I'd prefer if they were told to use `nightly` instead.

athei:
You could still tell developers to use nightly for local development. But the CI should use stable cause that is what we are delivering to our users.

dockerfiles/ink-ci-linux/Dockerfile
```diff
@@ -67,13 +67,15 @@ RUN	set -eux; \
...
+ 	target add wasm32-unknown-unknown --toolchain stable && \
+ 	rustup component add rust-src --toolchain stable && \
 	rustup toolchain install nightly-2022-07-28 --target wasm32-unknown-unknown \
 		--profile minimal --component rustfmt clippy miri rust-src rustc-dev llvm-tools-preview && \
 # Alias pinned toolchain as nightly, otherwise it appears as though we
 # don't have a nightly toolchain (i.e rustc +nightly --version is empty)
 	ln -s /usr/local/rustup/toolchains/nightly-2022-07-28-x86_64-unknown-linux-gnu \
-		/usr/local/rustup/toolchains/nightly-x86_64-unknown-linux-gnu && \
-	rustup default nightly && \
+ 	   	/usr/local/rustup/toolchains/nightly-x86_64-unknown-linux-gnu && \
```
athei, line 77: Docs are stale now.

dockerfiles/contracts-ci-linux/Dockerfile
```diff
@@ -42,13 +42,15 @@ RUN set -eux; \
...
+ 	target add wasm32-unknown-unknown --toolchain stable && \
+ 	rustup component add rust-src --toolchain stable && \
 	rustup toolchain install nightly-2022-07-28 --target wasm32-unknown-unknown \
 		--profile minimal --component rustfmt clippy rust-src && \
 	# Alias pinned toolchain as nightly, otherwise it appears as though we
 	# don't have a nightly toolchain (i.e rustc +nightly --version is empty)
 	ln -s /usr/local/rustup/toolchains/nightly-2022-07-28-x86_64-unknown-linux-gnu \
  	   	/usr/local/rustup/toolchains/nightly-x86_64-unknown-linux-gnu && \
-	rustup default nightly && \
+	rustup default stable && \
```
athei, line 53: Docs stale.

athei (approved):

PR #460, opened 2022-09-01
Pin nightly to updated version

cmichi:


dockerfiles/ink-ci-linux/Dockerfile
```diff
@@ -63,10 +63,18 @@ RUN set -eux; \
...
+	# Cargo Clippy, and Miri.
+	#
+	# Note that we pin the nightly toolchain since it often creates breaking changes during
+	# the RustFmt and Clippy stages of the CI.
+	rustup toolchain install nightly-2022-08-26 --target wasm32-unknown-unknown \
 		--profile minimal --component rustfmt clippy miri rust-src rustc-dev llvm-tools-preview && \
 
+	# Alias pinned toolchain as nightly, otherwise it appears as though we
+	# don't have a nightly toolchain (i.e rustc +nightly --version is empty)
+	ln -s /usr/local/rustup/toolchains/nightly-2022-08-26-x86_64-unknown-linux-gnu \
```
ascjones, line 75: Could use a variable for the nightly version, I can foresee someone changing one and not the other. Sure the docker build would fail but would be annoying.

ascjones (approved):

dockerfiles/ink-ci-linux/Dockerfile
cmichi: I agree, food for a follow-up.

PR #462, opened 2022-10-05
Kubectl Image

lazam:
Minimal image of kubectl

### DIFF
--- a/dockerfiles/kubetools/kubectl-1.22.Dockerfile
+++ b/dockerfiles/kubetools/kubectl-1.22.Dockerfile
@@ -1,0 +1,35 @@
+FROM docker.io/library/alpine:latest
+
+ARG VCS_REF=master
+ARG BUILD_DATE=""
+ARG REGISTRY_PATH=docker.io/paritytech
+ARG KUBE_VERSION="1.22.15"
+
+# metadata
+LABEL io.parity.image.authors="devops-team@parity.io" \
+    io.parity.image.vendor="Parity Technologies" \
+    io.parity.image.title="${REGISTRY_PATH}/kubetools" \
+    io.parity.image.description="ca-certificates git jq make curl gettext kubectl;" \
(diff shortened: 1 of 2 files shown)


dockerfiles/kubetools/kubectl-1.22.Dockerfile
```diff
@@ -0,0 +1,35 @@
+FROM docker.io/library/alpine:latest
+
+ARG VCS_REF=master
+ARG BUILD_DATE=""
+ARG REGISTRY_PATH=docker.io/paritytech
+ARG KUBE_VERSION="1.22.15"
```
bakhtin, line 6: I would suggest to not include the version in the file name. It will be a copy-paste of the file with only difference being the `kubectl` version when we need to bump it. Instead, we could put the `kubectl` version in the image tag to make different versions available (e.g., `docker.io/paritytech/kubetools-kubectl-1.22`)

BulatSaif:
Why you need separate image, I think we can update the main Dockerfile. 
We also should merge Dockerfile and helm3.Dockerfile (I didn't see what we used helm2)

lazam:
> Why you need separate image, I think we can update the main Dockerfile. We also should merge Dockerfile and helm3.Dockerfile (I didn't see what we used helm2)

We can use helm3 since it has kubectl, but if we want to have a minimal image with less package then it's better to have a dedicated image for it. We might include other packages in the future that might not be usable in our CI but beneficial in the node helm chart. 

I don't know if there's still a project that uses helm2 or the old dockerfile.

dockerfiles/kubetools/kubectl-1.22.Dockerfile
lazam: Another option is properly separate them in a directory under kubetools. It will look like this:

```
kubetools/
|--- kubectl-1.22/
|-- helm3/
```

BulatSaif (approved):

PR #464, opened 2022-10-05
fix multisig-ci

alvicsam:


oleg-plakida (approved):

dockerfiles/multisig-ci/Dockerfile
```diff
@@ -8,19 +8,18 @@ LABEL io.parity.image.authors="devops-team@parity.io" \
...
-    # add non-root user
-    groupadd -g 10000 nonroot && \
-    useradd -u 10000 -g 10000 -s /bin/bash -m nonroot
+	chmod +x kubectl && \
+	mv kubectl /usr/local/bin/ && \
+	apt-get update && apt-get --no-install-recommends install -y gettext python3-setuptools && \
+	pip install poetry && \
+	# add non-root user
+	groupadd -g 10000 nonroot && \
+	useradd -u 10000 -g 10000 -s /bin/bash -m nonroot
```
radupopa2010, line 23: I would use long arg to increase readability and bump the user ID to 10000 :) 

```
# UIDs below 10,000 are a security risk, as a container breakout could result
# in the container being ran as a more privileged user on the host kernel with
# the same UID.
#
# Static GID/UID is also useful for chown'ing files outside the container where


RUN groupadd --gid 10000 nonroot && \
    useradd  --home-dir /home/nonroot \
             --create-home \
             --shell /bin/bash \
             --gid nonroot \
             --groups nonroot \
             --uid 10000 nonroot

```

PR #467, opened 2022-10-06
update terraform version

jam206:
To build new terraform docker image

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -362,7 +362,7 @@ terraform:
   variables:
     <<:                            *default-vars
     # https://releases.hashicorp.com/terraform/
-    TERRAFORM_VERSION:             "1.0.4"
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -362,7 +362,7 @@ terraform:
   variables:
     <<:                            *default-vars
     # https://releases.hashicorp.com/terraform/
-    TERRAFORM_VERSION:             "1.0.4"
+    TERRAFORM_VERSION:             "1.1.0"
```
BulatSaif, line 365: Why not latest `1.3.1`?

jam206: I need an image for 1.1.0 for a specific deployment, after I also need to build 1.2.4+

jam206: Actually my code does work on 1.3.1, have changed to that.

PierreBesson (approved):

PR #488, opened 2022-11-23
Use `clang-14` for `paritytech/base-ci-linux`

rcny:
We have linker failures in the Substrate CI most possibly related that `clang-13` is too old for currently used Rust versions. This PR bumps `clang` from `13` to `14` for the base image and the dependent `ci-linux` image.

### DIFF
--- a/dockerfiles/base-ci-linux/Dockerfile
+++ b/dockerfiles/base-ci-linux/Dockerfile
@@ -31,8 +31,8 @@ COPY utility/base-ci-linux-config /root/.cargo/config
 ENV RUSTUP_HOME=/usr/local/rustup \
 	CARGO_HOME=/usr/local/cargo \
 	PATH=/usr/local/cargo/bin:$PATH \
-		CC=clang-13 \
-		CXX=clang-13
+		CC=clang-14 \
+		CXX=clang-14
 
 # install tools and dependencies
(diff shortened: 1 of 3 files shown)


altaua (approved):

altaua:
Hmm, several ci jobs had just failed due to transient network failures and I approved this pr before they'd all finished running again, but now it turns out several of them are actually legit failing, looks like some downloaded files are getting truncated

altaua:
Oh I see the *_VERSION flags aren't being passed in for some reason.

altaua (approved):
approving again, aiui now this is going to work when merged, it's just the *-pr jobs that don't set the *_VERSION variables

alvicsam:
About CI - I added the jobs that work in PR recently, so there might be some bugs

muharem:
@alvicsam nether it does not help, or my jobs running on the prev ci script version
https://gitlab.parity.io/parity/mirrors/substrate/-/jobs/2081682

samelamin:
Same thing for me

https://gitlab.parity.io/parity/mirrors/substrate/-/jobs/2081326

rcny:
We are unable to release the updated images due to the same linker error. Need more time to investigate, `clang-14` didn't help.

PR #490, opened 2022-11-24
Use `clang-15` for `paritytech/base-ci-linux`

rcny:
`clang-14` is still too low for the CI workloads even there were reports that it fixes related linker problems locally. Bumping further to `clang-15`.

cc @muharem @samelamin


alvicsam (approved):

rcny:
Supersedes #489.

samelamin:
Is it safe to rerun workloads now?

rcny:
Not yet, we're rebuilding the images, I'll post an update here when the we'll push updated images.

rcny:
So, looks like it's not related to the `clang` version as we continue to observe the same behavior with the same linker error even with `clang-15`. I will investigate more.

PR #491, opened 2022-11-24
Use LLVM's lld instead of GNU's ld

rcny:
`clang` used `ld` as a linker instead of `lld` whole this time. This PR fixes that and also downgrades LLVM and `clang` back to 14.

Superseeds #490 and relates to #488.


bkchr (approved):

rcny:
@muharem @samelamin Fixed for real this time. Thanks for the patience.

samelamin:
@rcny sorry to be a pain but it looks like there are still issues. But - at least from what I can see -  it's not related to the linker 

https://gitlab.parity.io/parity/mirrors/substrate/-/jobs/2084463

The `check-rustdoc` seems to be failing for all builds

rcny:
@samelamin Fix is in progress, could be tracked [here](https://github.com/paritytech/substrate/pull/12777).

PR #497, opened 2022-12-21
Add dependencies of publish-crates to ci-linux

joao-paulo-parity:
Adds the dependencies used in the crates publishing automation script (https://github.com/paritytech/releng-scripts/blob/master/publish-crates) to the `ci-linux` image so that the script doesn't have to install them by itself.

The automation is [currently used only in Substrate](https://github.com/paritytech/substrate/blob/47ebf8742144e1b18414ecf97f43c16de624a180/scripts/ci/gitlab/pipeline/publish.yml#L224) but it'll probably [also be used for other repositories in the future](https://github.com/paritytech/release-engineering/issues/141).


rcny:
Closing as stale.

PR #515, opened 2023-02-25
Update get-substrate.sh

ARC19-dev:
you can install dependencies package for Rust and Substrate framework with this script written script is useable for Redhat systems like Rocky linux, Alma linux, Centos and etc. Development Tools group use for build and complie system's projects;
 important package inside in Development Tools is:
  autoconf, automake, binutils, libtool, gcc, pkgconf, rpm-build and etc. 
openssl-devel: share object libraries for openssl
clang-devel: share object libraries for clang
protobuf: Protocol Buffers are Google's language-neutral, platform-neutral, extensible mechanism for serializing structured data
 we use for compile substrate framework becuse libp2p use protobuf libraries
protobuf-compiler: development tools for protobuf

### DIFF
--- a/get-substrate.sh
+++ b/get-substrate.sh
@@ -11,9 +11,9 @@ if [[ "$OSTYPE" == "linux-gnu" ]]; then
 
 	if [ -f /etc/redhat-release ]; then
 		echo "Redhat Linux detected."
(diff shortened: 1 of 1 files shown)


rcny:
Thanks!

rcny (approved):

PR #533, opened 2023-03-21
ws-health-exporter was improved

kogeler:
* added new environment variables to setup existing and new preferences.
* added the list of WebSocket URLs feature and the `url` metric tag.
* added the block drift check. it is disabled by default.
* logging preferences were refactored. It allows to use common logger for all modules.
* added the production HTTP server
* README.md was added

### DIFF
--- a/dockerfiles/ws-health-exporter/exporter.py
+++ b/dockerfiles/ws-health-exporter/exporter.py
@@ -65,7 +65,7 @@ def run_error(msg):
     sys.exit(1)
 
 
-def pars_config(config):
+def parse_config(config):
     env = Env()
     # WSHE_LOG_LEVEL
     # it checks debug CLI flag to keep backward compatibility with old versions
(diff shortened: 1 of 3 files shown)


dockerfiles/ws-health-exporter/exporter.py
```diff
@@ -2,54 +2,157 @@
...
+def uri_validator(url):
+    try:
+        result = urlparse(url)
+        return all([result.scheme, result.netloc])
+    except:
+        return False
+
+
+def run_error(msg):
+    print('fatal run error! ' + msg, file=sys.stderr)
```
BulatSaif, line 64: why not `logging.error`? Is not initialized yet? 
```suggestion
    logging.error('fatal run error! ' + msg, file=sys.stderr)
```

BulatSaif (approved):
Looks excelled!

dockerfiles/ws-health-exporter/exporter.py
```diff
@@ -2,54 +2,157 @@
...
+    sys.exit(1)
 
-def check_ws(node_url, timeout):
+
+def pars_config(config):
```
bakhtin, line 68:
```suggestion
def parse_config(config):
```

bakhtin (approved):

PR #539, opened 2023-03-30
migrate all base-ci consumers to base-ci-linux

altaua:


alvicsam (approved):

rcny (approved):

alvicsam (changes requested):
contracts-ci-linux is still in use

PR #545, opened 2023-04-12
feat(ci-linux): tag staging images with rust & debian version

altaua:
resolves paritytech/ci_cd/issues/776

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -332,7 +332,30 @@ base-ci-linux:
 ci-linux:
   <<:                              *docker_build
   script:
-    - *push_to_staging
+    - buildah bud
+        --format=docker
+        --build-arg VCS_REF="$CI_COMMIT_SHA"
+        --build-arg BUILD_DATE="$(date +%Y%m%d)"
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -332,7 +332,28 @@ base-ci-linux:
...
   script:
-    - *push_to_staging
+    - buildah bud
+        --format=docker
+        --build-arg VCS_REF="$CI_COMMIT_SHA"
+        --build-arg BUILD_DATE="$(date +%Y%m%d)"
+        --build-arg REGISTRY_PATH="$REGISTRY_PATH"
+        --tag "$REGISTRY_PATH/$IMAGE_NAME:staging"
+        --file "dockerfiles/$IMAGE_NAME/Dockerfile" dockerfiles
+
```
alvicsam, line 342: Could you add comment with what we expect in CONTAINER_ID and VERSION_TAG?

altaua: I've added some comments to clarify. CONTAINER_ID is just an internal buildah container id (akin to what docker run would print), so the exact format of that doesn't matter to us.

PR #555, opened 2023-05-18
Introduce unified CI image

rcny:
I want to try to gradually replace all our CI images with a single one that is built from the single source of truth, tagged as `<distro codename>-<rust version>`, plus an additional `<distro codename>` tag. I don't think that image size matters, if we'll also push the image to our internal registry located closely to the CI runners or just start proxying the Docker Hub upstream.

What this PR introduces right now:
* New `ci-unified` image that combines `base-ci-linux` and `ci-linux` (will absorb other images later)
* Bump the base image to Ubuntu 22.04 (needed for my Firebuild tests, could be downgraded back to `stretch` if there will be no success with Firebuild)
* Bump LLVM to 15 and allow upping its version via a single change to the relevant `ARG`
* Pin stable Rust to bump it explicitly in the future cases when needed
* Use consistent Dockerfile formatting
* Build the image on beefy `docker+machine` runners


sergejparity (approved):
Cool

PR #561, opened 2023-05-19
debian 11 docker image was added

kogeler:
The debian 11 docker image was added because we need it to run new versions of node binaries.


dockerfiles/debian11/README.md
```diff
@@ -0,0 +1,4 @@
+# Description
+
+This Debian 10 image is used in the ansible Molecule test as a base to apply roles.
```
BulatSaif, line 3:
```suggestion
This Debian 11 image is used in the ansible Molecule test as a base to apply roles.
```

rcny (approved):

PR #570, opened 2023-06-09
Downgrade rust nightly version

kalaninja:
There is a [known problem](https://stackoverflow.com/questions/75955457/substrate-node-template-cannot-create-a-runtime-error-othercannot-deserialize) with wasm compilation on the latest rust nightly. For example, [this job](https://gitlab.parity.io/parity/mirrors/trappist/-/jobs/2959599) currently fails with `cannot deserialize module: UnknownOpcode(192)`. 

Version `2023-01-01` is known to work.

### DIFF
--- a/dockerfiles/ci-linux/Dockerfile
+++ b/dockerfiles/ci-linux/Dockerfile
@@ -4,7 +4,7 @@ ARG REGISTRY_PATH=docker.io/paritytech
 
 FROM ${REGISTRY_PATH}/base-ci-linux:latest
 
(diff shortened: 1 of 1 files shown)


oleg-plakida (approved):

rcny:
Should not break the `cargo +nightly fmt`, but we'll see.

altaua:
> Should not break the `cargo +nightly fmt`, but we'll see.

spoiler alert: it did break substrate's fmt job

PR #571, opened 2023-06-19
feat(ci-linux): bundle second nightly just for rustfmt

altaua:
This way we can use different nightlies for rustfmt and wasm compilation.


altaua:
Closing in favour of #572

PR #591, opened 2023-08-15
Attempt to install clippy for bridges repo

bkontur:
trying to solve: https://gitlab.parity.io/parity/mirrors/parity-bridges-common/-/jobs/3389408


rcny:
Is behavior the same right now with the `ci-unified` image?

PR #592, opened 2023-08-29
More logs, block rate for ws-health-exporter

kogeler:
* if you check container logs you can't understand which check failed. new info logs help with it
* rate of new blocks feature was added. it is disabled by default. it can be useful in some cases

### DIFF
--- a/dockerfiles/ws-health-exporter/README.md
+++ b/dockerfiles/ws-health-exporter/README.md
@@ -18,9 +18,10 @@ It can be configured using environment variables:
   default `ws://127.0.0.1:5556`
 * `WSHE_NODE_MAX_UNSYNCHRONIZED_BLOCK_DRIFT` - setups maximum of unsynchronized blocks; if a node has
   more, the health check will fail; default `0` blocks (disabled)
-* `WSHE_NODE_MIN_PEERS` - setups minimum of peers; if a node has
-  less, the health check will fail; default `10` peers
-* `WSHE_BLOCK_RATE_MEASUREMENT_PERIOD` - average rate of new blocks is calculated for the period; default `600` seconds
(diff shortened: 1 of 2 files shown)


BulatSaif (approved):

dockerfiles/ws-health-exporter/README.md
```diff
@@ -17,9 +17,13 @@ It can be configured using environment variables:
...
 * `WSHE_NODE_MAX_UNSYNCHRONIZED_BLOCK_DRIFT` - setups maximum of unsynchronized blocks; if a node has
-  more, the health check will be failed; default `0` (disabled)
-* `WSHE_NODE_MIN_PEERS` - setups minimum of peers; if a node has
-  less, the health check will be failed; default `10`
+  more, the health check will fail; default `0` blocks (disabled)
+* `WSHE_NODE_MIN_PEERS` - setups minimum of peers; if a node has less, the health check will fail; default `10` peers
+* `WSHE_BLOCK_RATE_MEASUREMENT_PERIOD` - average rate of new blocks is calculated for the period;
+  if this period is less than `WSHE_WS_CHECK_INTERVAL` it will be equal `WSHE_WS_CHECK_INTERVAL`; 
+  default `600` seconds
+* `WSHE_MIN_BLOCK_RATE` - setups minimum rate of new blocks; if a node has a
```
PierreBesson, line 25: I don't think it is a good idea to have this check. This assumes the chain will be producing blocks no matter what which is far from a guaranteed assumption. So this is not necessarily indicative of a failure of the node and it would be quite bad to have this used for load-balancing in case the chain slows down below this min_block_rate threshold

kogeler: Yes. But it can be useful for some production validators for example, especially if you restrict the maximum number of restart attempts.

Anyway it is disabled by default.

PR #595, opened 2023-09-05
Add new container image for prdoc

chevdor:
New container for https://github.com/paritytech/prdoc

This needs:
- [x] cleanup (@chevdor)
- [ ] a scheduled build. I think a weekly build + option for a manual trigger would be enough.
- [ ] a first build

cc @alvicsam @altaua

### DIFF
--- a/dockerfiles/prdoc/.dockerignore
+++ b/dockerfiles/prdoc/.dockerignore
@@ -1,0 +1,7 @@
+.env*
+!.env-sample
+tests
+.github
+.git
+.*
+CODEOWNERS
--- a/dockerfiles/prdoc/Dockerfile
+++ b/dockerfiles/prdoc/Dockerfile
@@ -1,0 +1,41 @@
+FROM docker.io/paritytech/ci-linux:production as builder
+
+ARG PROFILE=release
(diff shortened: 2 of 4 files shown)


alvicsam:
I would prefer not merging this and create a CI for `prdoc` repo that will build the image

chevdor:
@alvicsam I think we need to weight your reasons vs security aspects. Could you explain why you'd rather keep the container out of this repo ?

alvicsam:
Initially the repo was for infra images, they don't require any code to be built. If `prdoc` was a bunch of few script and a `dockerfile` then it would make sense to put it there. When the repo has a lot of files it makes more sense to build the image from the repo on tag or push to master.
Also here we have a common CI for all files so we won't use your script (`dockerfiles/prdoc/build.sh`).
Security aspects I'd prefer to discuss in matrix.

chevdor:
> Also here we have a common CI for all files so we won't use your script (dockerfiles/prdoc/build.sh).

Yes, that's fine, I provide one more as a way to test/build locally.

> Security aspects I'd prefer to discuss in matrix.

Ok, let's do that.

PR #596, opened 2023-09-05
Add container image for ruled_labels

chevdor:
Same than #595, just another binary.


chevdor:
Closing for now since we have an image in the project's repo.

PR #601, opened 2023-09-22
Add --locked to estuary install to fix build

ascjones:
Docker image build fails installing `estuary`, adding `--locked` should fix it.


ascjones:
Same fix is in #602

PR #615, opened 2023-12-11
Update rve toolchain in unified image

pgherveou:
Update the unified-ci script with the latest version of the risc-v toolchain and openssl to version 3.1.4

### DIFF
--- a/dockerfiles/ci-unified/Dockerfile
+++ b/dockerfiles/ci-unified/Dockerfile
@@ -104,10 +104,10 @@ RUN ln -s "/usr/local/rustup/toolchains/nightly-${RUST_NIGHTLY_VERSION}-x86_64-u
 RUN rustup target add wasm32-unknown-unknown
 
 # Install rve-nightly-2023-04-05
-RUN wget https://github.com/paritytech/rustc-rv32e-toolchain/releases/download/rust-rve-nightly-2023-04-05-r0-x86_64-unknown-linux-gnu/rust-rve-nightly-2023-04-05-r0-x86_64-unknown-linux-gnu.tar.zst && \
-    tar --zstd -xf rust-rve-nightly-2023-04-05-r0-x86_64-unknown-linux-gnu.tar.zst && \
-    cd rust-rve-nightly-2023-04-05-r0-x86_64-unknown-linux-gnu && \
-    mv tmp/rve-nightly ~/.rustup/toolchains && \
+
+RUN wget https://github.com/paritytech/rustc-rv32e-toolchain/releases/download/rust-rve-nightly-2023-04-05-x86_64-unknown-linux-gnu/rust-rve-nightly-2023-04-05-x86_64-unknown-linux-gnu.tar.zst && \
+    tar --zstd -xf rust-rve-nightly-2023-04-05-x86_64-unknown-linux-gnu.tar.zst && \
+    mv tmp/rve-nightly /usr/local/rustup/toolchains && \
     rm -f rust-rve-nightly-2023-04-05-r0-x86_64-unknown-linux-gnu.tar.zst && \
     rmdir tmp
 
--- a/dockerfiles/ci-unified/Dockerfile
+++ b/dockerfiles/ci-unified/Dockerfile
@@ -104,13 +104,12 @@ RUN ln -s "/usr/local/rustup/toolchains/nightly-${RUST_NIGHTLY_VERSION}-x86_64-u
 RUN rustup target add wasm32-unknown-unknown
 
 # Install rve-nightly-2023-04-05
-RUN wget https://github.com/paritytech/rustc-rv32e-toolchain/releases/download/rust-rve-nightly-2023-04-05-r0-x86_64-unknown-linux-gnu/rust-rve-nightly-2023-04-05-r0-x86_64-unknown-linux-gnu.tar.xz && \
-    tar xfz rust-rve-nightly-2023-04-05-r0-x86_64-unknown-linux-gnu.tar.xz && \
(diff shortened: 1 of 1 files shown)


dockerfiles/ci-unified/Dockerfile
```diff
@@ -281,6 +275,18 @@ RUN	cargo install --git https://github.com/paritytech/ink-waterfall.git csv-comp
 # parity-scale-codec
 RUN	cargo +nightly-${RUST_NIGHTLY_VERSION} install grcov rust-covfix xargo dylint-link
 
+# Install rve-nightly-2023-04-05
+
```
pgherveou, line 279:
```suggestion
```

dockerfiles/ci-unified/Dockerfile
```diff
@@ -103,21 +103,15 @@ RUN ln -s "/usr/local/rustup/toolchains/nightly-${RUST_NIGHTLY_VERSION}-x86_64-u
...
 RUN apt-get install -y --no-install-recommends g++
 RUN rustup target add x86_64-unknown-linux-musl
 
 # generic ci | install common cargo tools
-RUN cargo install cargo-web wasm-pack cargo-deny cargo-spellcheck cargo-hack \
+RUN cargo install cargo-web wasm-pack cargo-deny cargo-hack \
                   mdbook mdbook-mermaid mdbook-linkcheck mdbook-graphviz mdbook-admonish mdbook-last-changed
 
-RUN cargo install cargo-nextest --locked
+RUN cargo install cargo-spellcheck cargo-nextest --locked
```
chevdor, line 115: It is good practise to avoid the multipliction of layers in a docker image.
So instead of:
```
RUN cargo install foo
RUN cargo install bar
```

the following is preferable:
```
RUN cargo install foo bar
```

If you prefer a mutiline option to keep things clear, you may use:
```
RUN cargo install \
            foo bar \
            baz
```

dockerfiles/ci-unified/Dockerfile
```diff
@@ -281,6 +275,18 @@ RUN	cargo install --git https://github.com/paritytech/ink-waterfall.git csv-comp
 # parity-scale-codec
 RUN	cargo +nightly-${RUST_NIGHTLY_VERSION} install grcov rust-covfix xargo dylint-link
 
+# Install rve-nightly-2023-04-05
+
+RUN wget https://github.com/paritytech/rustc-rv32e-toolchain/releases/download/rust-rve-nightly-2023-04-05-x86_64-unknown-linux-gnu/rust-rve-nightly-2023-04-05-x86_64-unknown-linux-gnu.tar.zst && \
+    tar --zstd -xf rust-rve-nightly-2023-04-05-x86_64-unknown-linux-gnu.tar.zst && \
+    mv tmp/rve-nightly /usr/local/rustup/toolchains && \
```
chevdor, line 282: It would be safer to use `/tmp` instead of `tmp`

dockerfiles/ci-unified/Dockerfile
```diff
@@ -103,21 +103,15 @@ RUN ln -s "/usr/local/rustup/toolchains/nightly-${RUST_NIGHTLY_VERSION}-x86_64-u
...
 # generic ci | install common cargo tools
-RUN cargo install cargo-web wasm-pack cargo-deny cargo-spellcheck cargo-hack \
+RUN cargo install cargo-web wasm-pack cargo-deny cargo-hack \
                   mdbook mdbook-mermaid mdbook-linkcheck mdbook-graphviz mdbook-admonish mdbook-last-changed
 
-RUN cargo install cargo-nextest --locked
+RUN cargo install cargo-spellcheck cargo-nextest --locked
 
 # generic ci | diener 0.4.6 | NOTE: before upgrading please test new version with companion build, example can be found here: https://github.com/paritytech/substrate/pull/12710
 RUN cargo install diener --version 0.4.6
```
chevdor, line 117: it would be good to pack all the cargo install together if possible

chevdor (approved):
LGTM besides the small improvements for the Container image

alvicsam:
@chevdor is it okay that openssl will be upgraded in ci image to 3.1.4 version?

chevdor:
@alvicsam I don't know which version is in the image right now, assuming the new version does not involve breaking changes, it will be fine. I am not aware of anything depending on a specific version.

We will see that in CI way before we need to use it in the release anyway. So it is safe and wise to try updating to a current version.

alvicsam:
@chevdor It is needed for adding risc-v toolchain to the ci needed. It requires a newer openssl version. Current version is 1.1.sth and with this PR openssl will be installed from `unstable` and updated to 3.1.4 version. The newer openssl version was tested in [this PR](https://github.com/paritytech/polkadot-sdk/pull/2554/commits/ddbddaffe52a265f9c4c31dcf6d1bb73fd2da839) and afaiu works well.

dockerfiles/ci-unified/Dockerfile
pgherveou: Not sure why we don't do --locked for all of them. I assumed it was to save some downloads and re-use as many common crates as possible

dockerfiles/ci-unified/Dockerfile
pgherveou: it's ./tmp here that is extracted from the tarball

dockerfiles/ci-unified/Dockerfile
pgherveou: just changed what was breaking, I think we should move all these install to use --locked.
Otherwise this could break for future version as I discovered here.

PR #640, opened 2024-05-30
`ci-unified` "fixed" versions and layers cache

AndWeHaveAPlan:
Move rust, forklift, codecov versions variables to separate file to force updates only by commit only
Slightly changed image tag timestamp suffix to (%Y%m%d -> %Y%m%d:%H%M)
Add ci-unified layers caching


alvicsam (approved):

dockerfiles/ci-unified/Dockerfile
```diff
@@ -3,16 +3,11 @@ FROM docker.io/library/debian:bullseye-20230522-slim
 
 ### meta ###
```
alvicsam, line 5:
```suggestion
# Static vars, dynamic vars can be found in the build-args file. Variables were moved for caching.
```

alvicsam (approved):
lgtm, i'd also add a comment about variables

PR #647, opened 2024-07-22
Update rust toolchain to 1.79.0

serban300:
Needed for this parity-scale-codec PR: https://github.com/paritytech/parity-scale-codec/pull/615 which uses inline const.


alvicsam:
Superseded  by https://github.com/paritytech/scripts/pull/648

PR #651, opened 2024-08-14
Adding `cargo-edit` and `gh` to ci-unified

mutantcornholio:
This allows Cargo.toml manipulation in CI, primarily, versioning.
Will be useful for publishing crates, based on tags.


dockerfiles/ci-unified/Dockerfile
```diff
@@ -83,6 +83,9 @@ RUN pip install yq
 
 ### generic ci ####
 
+RUN wget https://github.com/cli/cli/releases/download/v2.54.0/gh_2.54.0_linux_amd64.deb && \
```
alvicsam, line 86: I think it worth moving the `2.54.0` version to an ARG

alvicsam (approved):

PR #652, opened 2024-08-16
Adding /__w directory to tools image

mutantcornholio:
This (in theory) will allow running GHA in container, without root

see https://github.com/actions/checkout/issues/1014


mutantcornholio:
tried with a custom image; didn't work

